PR #44, opened 2016-05-23
Fix orientation overwriting

aaronzufall:
Fixes an [issue](https://github.com/shogo4405/lf.swift/issues/43) where the AVMixer orientation is overwritten to current device orientation during publishing start


shogo4405:
This line is necessary for syncOrientation mode.

PR #150, opened 2016-12-15
Fix for unstop Audio Session

denizak:
Somehow when rtmpConnection close connection, it still keep reside in memory.


shogo4405:
> it still keep reside in memory.

This is an issue which cannot be ignored. But It seems that will stop video preview.

PR #157, opened 2016-12-30
Remove Target Remembership to fix Carthage Build Error: /.../Carthage…

anotheren:
Remove Target Remembership to fix Carthage Build Error: warning: no rule to process file '…/Checkouts/lf.swift/Examples/iOS/Screencast/README.md' of type net.daringfireball.markdown for architecture armv7/arm64...

Please create a new tag as the `tag 0.5.5` is not correctly work with Carthage, Thank you.


shogo4405:
> Please create a new tag as the tag 0.5.5 is not correctly work with Carthage,
Yes. I will do 3rd Jan 2017.

Issue #138, opened 2016-11-16
localRecord performance


pzs7602:
when publishing stream with publish type: .localRecord, the local recorded video's FPS never reaches 30(generally 6-8 FPS with iPhone 6 plus or iPhone 7 plus), the local video size is 
1280*720 or 1920*1080:
```
        rtmpStream.captureSettings = [
            "sessionPreset": AVCaptureSessionPreset1280x720,
            "continuousAutofocus": true,
            "continuousExposure": true,
        ]

        rtmpStream.videoSettings = [
            "width": 1280,
            "height": 720,
        ]
        rtmpStream.recorderSettings = [
            AVMediaTypeAudio: [
                AVFormatIDKey: Int(kAudioFormatMPEG4AAC),
                AVSampleRateKey: sampleRate,
                AVNumberOfChannelsKey: 2,
            ],
            AVMediaTypeVideo: [
                AVVideoCodecKey: AVVideoCodecH264,
                AVVideoHeightKey: 0,
                AVVideoWidthKey: 0,
            ],
            
        ]
```
while debugging, the device's CPU utilization is about 40~50%, so optimization may be possible for 30FPS local recorded video.

buubui:
I also have an issue with localRecord, the fps is very low so that the video is unusable. The setting keys are messy for me.  Can someone please give me an example of configuration to increase the fps (reduce the video quality a bit is fine )? Thanks.

shogo4405:
6924c9f @anotheren Thank u.

Benchmark.

| |Stream FPS|Video FPS|Parameter|
|:-:|:-:|:-:|:-:|
|iPhone7 Plus|30fps|30fps| AVCaptureSessionPreset1280x720, stream={w=1280, h=720}, record={w=0, h=0} |
|iPod touch 5th|15-16fps|15-16fps| AVCaptureSessionPreset1280x720, stream={w=1280}, h=720, record={w=0, h=0}|

PR #160, opened 2017-01-05
Fix Issues #138, now record video will have full FPS

anotheren:
Fix Issues #138, now record video will have full FPS

Fix record video quality when add effects


shogo4405:
Thanks a lot!

PR #196, opened 2017-03-05
Fix crashing on ios

ra1028:
Hi, @shogo4405 
Thanks for the awesome lib!
I found and patched some crashing bugs below.

```
1. ProgramSpecific.swift
  Cause: Bad access to array via concurrent thread
  Patch: Thread lock using mutex

2. TSWriter.swift
  Cause: Raised an exception when invoke FileHandle.synchronizeFile function
  Patch: Catch exception by nstry

3. HTTPRequest.swift
  Cause: Raised index out of range exception
  Patch: Add guard statement before take values from bytes string
```

Please confirm this :)

### DIFF
--- a/Sources/HTTP/HTTPRequest.swift
+++ b/Sources/HTTP/HTTPRequest.swift
@@ -36,7 +36,11 @@ extension HTTPRequestCompatible {
                     break
                 }
             }
-            let first:[String] = lines.first!.components(separatedBy: " ")
+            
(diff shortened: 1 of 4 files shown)


ra1028:
@shogo4405 
Would you explain me why close this PR?

shogo4405:
@ra1028 
Sorry. Operation mistake...

ra1028:
Ok :)
Thanks for merging!

PR #214, opened 2017-04-19
Add a flag to disable saving to photo library on iOS

akolov:
Would you please consider a flag to disable writing local recording to a photo library? 
Sometimes it's useful for logic internal to the app and doesn't need to appear in photo library.


shogo4405:
OK. Please use this.
(rtmpStream.mixer.recorder.delegate as? DefaultAVMixerRecorderDelegate)?.shouldSaveToPhotoLibrary = false

PR #235, opened 2017-06-04
red5+rtmpt unlock video encoder

zhuker:
unlocks video encoder from rtmpt socket

https://github.com/shogo4405/lf.swift/issues/233

### DIFF
--- a/Sources/RTMP/RTMPTSocket.swift
+++ b/Sources/RTMP/RTMPTSocket.swift
@@ -89,6 +89,9 @@ final class RTMPTSocket: NSObject, RTMPSocketCompatible {
         for chunk in chunks {
             bytes.append(contentsOf: chunk)
         }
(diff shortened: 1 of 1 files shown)


Sources/RTMP/RTMPTSocket.swift
```diff
@@ -89,6 +89,9 @@ final class RTMPTSocket: NSObject, RTMPSocketCompatible {
         for chunk in chunks {
             bytes.append(contentsOf: chunk)
         }
+        if (locked != nil) {
+            OSAtomicAnd32Barrier(0, locked!)
```
zhuker: @shogo4405 same as in `NetSocket.swift` unlocks video encoder not sure this is the right place to put it though

shogo4405:
I will test a a few days later :)

PR #242, opened 2017-06-09
rtmpts

zhuker:
like https://www.ffmpeg.org/ffmpeg-protocols.html#rtmpts


zhuker:
@shogo4405 rtmpt over https tunnel support

PR #243, opened 2017-06-09
supportedProtocols rtmpts

zhuker:


zhuker:
@shogo4405 sorry one more commit

PR #258, opened 2017-07-15
Retrying the request if the keepalive connection failed (fixes #256)

### LINKED ISSUE #256: RTMPTSocket reconnect?
server i am sending my rtmpt requests to (nginx) has a request count limit after limit is exceeded http keep-alive connection is being dropped. lf.swift treats it as connection was closed and stops streaming. 
can you point me to the best place to look into it?
@shogo4405

geeee:


Sources/RTMP/RTMPTSocket.swift
```diff
@@ -129,9 +132,22 @@ final class RTMPTSocket: NSObject, RTMPSocketCompatible {
...
+                    logger.verbose("Will retry request for path=\(lastRequestPathComponent)")
+                }
+                isRetryingRequest = true
+                doRequest(lastRequestPathComponent, lastRequestData, listen)
+            }
+
             return
         }
 
+        isRetryingRequest = false
```
zhuker: Should be synchronized? (Like output queue sync)

PR #269, opened 2017-07-31
make logger optional

pvinis:
dont depend on xcglogger. everyone can use their choice of logger.


shogo4405:
> everyone can use their choice of logger.

Yes. I'm think so. And I'm considering another approch.

pvinis:
ok.
what are you thinking?
can i help?

PR #270, opened 2017-07-31
Start recording when needed

pvinis:
i want to be able to start streaming first, and start recording later.

### DIFF
--- a/Examples/iOS/LiveViewController.swift
+++ b/Examples/iOS/LiveViewController.swift
@@ -123,7 +123,8 @@ final class LiveViewController: UIViewController {
         if let data:ASObject = e.data as? ASObject , let code:String = data["code"] as? String {
             switch code {
             case RTMPConnection.Code.connectSuccess.rawValue:
-                rtmpStream!.publish(Preference.defaultInstance.streamName!)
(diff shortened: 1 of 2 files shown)


shogo4405:
Wowza(FMS) can server side recording. publish("foo", "record" | "append" | "appendWithGap")

You can start recording.
publish("foo") then publish("foo", .localRecord)

I do not want to increase methods.

pvinis:
I know wowza can do stuff. this is for local recording. I didn't know I can publish twice. then I will try it. but it's it going to be slower/harder on the phone? will the frame processing happen twice with two publish?

shogo4405:
I designed...
https://github.com/shogo4405/HaishinKit.swift/blob/master/Sources/RTMP/RTMPStream.swift#L422-L432

> but it's it going to be slower/harder on the phone
It is a bug. I will try fix.

pvinis:
ok thanks for the answer. I will try it and then maybe I give a PR with some refactoring so you check it too. I want to help you make this lib better :)

purvangi-dev:
Is recording in between live streaming working? Can you share working samples please?

PR #337, opened 2017-11-09
fixed crash in AudioConverterFillComplexBuffer

### LINKED ISSUE #320: AV out of sync after camera toggle
If I flip back/front camera with the code below, the video is a fragment of second behind audio. This seems to be cumulative (lag increases each toggle)

```
 func flipCamera() {
        let position:AVCaptureDevice.Position = currentPosition == .back ? .front : .back
        rtmpStream.attachCamera(DeviceUtil.device(withPosition: position)) { error in
            //logger.warning(error)
            log_error("attach camera error: \(error)")
        }
        currentPosition = position
    }
```

macness-pavelh:
Hi I believe this fixes random crashes in AudioConverterFillComplexBuffer mentioned in issues some time back. 

I believe the logic of AACEncoder.encodeSampleBuffer is not perfect. See for example the last paragraph here: https://developer.apple.com/library/content/qa/qa1317/_index.html - calling AudioConverterFillComplexBuffer can not assume that all data will be produced in a single shot.

Calling AudioConverterFillComplexBuffer does not ensure all data from converted to be consumed so we need to call it in a loop to ensure that as much as possible data was consumed from the encoder.

I'm not 100% sure about return codes, this is but a good and working attempt to fix it. You should check if the pulling loot will always terminate (always gets status <= 0)

As a side-effect this fixes #320 too.

### DIFF
--- a/Sources/Codec/AACEncoder.swift
+++ b/Sources/Codec/AACEncoder.swift
@@ -188,37 +188,42 @@ final class AACEncoder: NSObject {
             }
         }
 
-        var ioOutputDataPacketSize:UInt32 = 1
-        let dataLength:Int = blockBuffer!.dataLength
-        let outOutputData:UnsafeMutableAudioBufferListPointer = AudioBufferList.allocate(maximumBuffers: 1)
-        outOutputData[0].mNumberChannels = inDestinationFormat.mChannelsPerFrame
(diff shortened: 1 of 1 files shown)


shogo4405:
Thank u!

PR #338, opened 2017-11-09
fix bitrate information

tue-savvy:
Fix RTMP metadata bitrate


shogo4405:
Please don't include 761d34f commit.

PR #377, opened 2018-02-27
Remove extraneous space from auth command url.

milesegan:
Because of the space in this url string the URL construction always fails and the ! operator results in a crash.


shogo4405:
orz.. Thank u!

PR #381, opened 2018-03-07
fixed autofocus after focussing point of interest

GerritPost:
After using setPointOfInterest to focus, autofocus is not working anymore. 
by changing the focus- and exposuremode to continuousAuto it now works just like native ios camera.

It will focus at a selected point, but when you move the camera, autofocus takes over.


shogo4405:
Thank u.

PR #395, opened 2018-03-28
Support for custom RTMPSocketCompatible implementations

gegles:
As part of the project I work on we needed to be able to provide RTMPConnection with our custom RTMPSocketCompatible class (proprietary protocol).

In order for this to be possible, I defined an RTMPSocketFactory protocol and reproduced the existing logic via a DefaultRTMPSocketFactory.

I also had to make a couple of classes public (RTMPChunk, RTMPSocketDelegate).

Let me know how I can improve this if needed, any feedback is welcome.

Thanks. G.


shogo4405:
There is no need in many cases.
Please fork this project.

PR #400, opened 2018-04-01
Async set NetStream settings

ronhuang:
In commit b3d6698, various settings in NetStream were changed to be updated synchronously. However, this is causing deadlock when video settings are updated (e.g. in ``viewWillAppear()``) when returning back from another view controller. This commit changes ``sync`` back to ``async`` to avoid deadlock.

### DIFF
--- a/Sources/Net/NetStream.swift
+++ b/Sources/Net/NetStream.swift
@@ -74,7 +74,7 @@ open class NetStream: NSObject {
             return  audioSettings
         }
         set {
-            lockQueue.sync {
+            lockQueue.async {
                 self.mixer.audioIO.encoder.setValuesForKeys(newValue)
             }
         }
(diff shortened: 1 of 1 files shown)


shogo4405:
This commit contains side effects. I can't accept.

> this is causing deadlock when video settings are updated (e.g. in viewWillAppear())
Please write steps to reproduce code.

ronhuang:
I have modified the example iOS app to reproduce this issue. The changes are at ronhuang/HaishinKit.swift@fdc563b30e21ffa4790e0fbc5ad6f7ad476b510b.

Steps to reproduce:
1. Build example iOS app with the above modifications
2. Launch app (in simulator will be sufficient)
3. Switch to Second tab
4. Update any text field to change ``Preference.defaultInstance.bitrate``
5. Switch back and forth between First and Second tabs until UI hangs
6. Once UI hangs, click "Pause program execution" in Xcode
7. Exam the call stack and see thread is stuck at ``__ulock_wait`` (as illustrated in the following screenshot)

netstream-videosettings-deadlock

PR #402, opened 2018-04-01
Record content locally without streaming to remote server

ronhuang:
Implement LocalStream, which inherits NetStream, to support recording content locally without streaming to remote server.

Obviously this feature will not be useful when source is the built-in camera, as the built-in camera app already offers recording functionality. But sources other than the built-in camera, which sample buffers are appended directly via ``appendSampleBuffer()``, can still benefit from this.


shogo4405:
This feature is nice-to-have. But isn't need-to-have.
I can't accept this PR.

PR #405, opened 2018-04-04
Fix Swift 4.1 'Overlapping Accesses' Error

MikeDonahue:
I updated Xcode to 9.3, Swift 4.1 and ran into this new error on build. Creating a local copy like the error and these [Apple Docs](https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/MemorySafety.html) suggested fixes the error. 

screen shot 2018-04-04 at 11 04 51 am


shogo4405:
https://github.com/shogo4405/HaishinKit.swift/pull/396
Fixed it. Do you have other problems?

MikeDonahue:
Sorry did not see that. I'll close this now.

Issue #429, opened 2018-06-22
Random "dispatch_sync called on queue already owned by current thread" crash


psegalen:
My app randomly crashes when using HaishinKit, here is the stack:
```
Application Specific Information:
BUG IN CLIENT OF LIBDISPATCH: dispatch_sync called on queue already owned by current thread
...
Thread 11 name:  Dispatch queue: com.haishinkit.HaishinKit.NetStream.lock
Thread 11 Crashed:
0   libdispatch.dylib             	0x0000000186a430d0 _dispatch_sync_wait + 600
1   HaishinKit                    	0x0000000100f2c6f8 _T010HaishinKit9NetStreamC13audioSettingss10DictionaryVySSypGvgTm + 156
2   HaishinKit                    	0x0000000100f2c6f8 _T010HaishinKit9NetStreamC13audioSettingss10DictionaryVySSypGvgTm + 156
3   Seize                         	0x00000001008cdb38 specialized SZBroadcastView.observeValue(forKeyPath:of:change:context:) + 56120 (SZBroadcastView.swift:108)
4   Seize                         	0x00000001008c8850 @objc SZBroadcastView.observeValue(forKeyPath:of:change:context:) + 34896 (SZBroadcastView.swift:0)
5   Foundation                    	0x00000001879ab9f0 NSKeyValueNotifyObserver + 304
6   Foundation                    	0x00000001879ab518 NSKeyValueDidChange + 404
7   Foundation                    	0x0000000187a65324 -[NSObject+ 865060 (NSKeyValueObservingPrivate) _changeValueForKeys:count:maybeOldValuesDict:usingBlock:] + 668
8   Foundation                    	0x00000001879aa9b4 -[NSObject+ 100788 (NSKeyValueObservingPrivate) _changeValueForKey:key:key:usingBlock:] + 60
9   Foundation                    	0x0000000187a62b40 _NSSetUnsignedShortValueAndNotify + 272
10  HaishinKit                    	0x0000000100f546cc _T010HaishinKit10RTMPStreamC10readyStateAC05ReadyE0OvW + 542412 (RTMPStream.swift:258)
11  HaishinKit                    	0x0000000100f56880 closure #1 in RTMPStream.publish(_:type:) + 551040 (RTMPStream.swift:422)
12  HaishinKit                    	0x0000000100efff8c _T0Ieg_IeyB_TR + 196492 (SoundSpliter.swift:0)
[...]

PR #432, opened 2018-06-27
Avoid dispatch_sync deadlocks (#429)

psegalen:
Fixes issue #429 

Basically, #429 problem is caused by using `dispatch_sync` (`lockQueue.sync`) from the concerned queue, which can cause deadlocks, which crash the app running.
This PR aims to check which is current queue to use `lockQueue.sync` only if the code's not already running on the queue.

Half the code is @clmntcrl's, huge thanks to him for his help!

### DIFF
--- a/Sources/Net/NetStream.swift
+++ b/Sources/Net/NetStream.swift
@@ -15,7 +15,13 @@ protocol NetStreamDrawable: class {
 // MARK: -
 open class NetStream: NSObject {
     public private(set) var mixer: AVMixer = AVMixer()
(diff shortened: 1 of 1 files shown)


psegalen:
@shogo4405 I've published a Release build on TestFlight with HaishinKit compiled from source with this fix: it throws an EXC_BAD_ACCESS when on line `rtmpStream.togglePause()` in my app.
I think it's related to the fact that I build from source, I may have missed some configuration piece.
Something odd: everything works like expected in Debug builds.

shogo4405:
Thank u.

Issue #118, opened 2016-10-10
How to play publish RTMP stream on VLC or other player?


wasimsafdar:
How can I play RTMP live stream on VLC and flash player after publishing it on local network? I tried but it always fails.

My URL streamed link is as follows:
rtmp://[local-IPAddress]/trainer/live1/streamName

wasimsafdar:
If I use RTMP server and online player, then I am able to stream. For example, if  url mentioned in following online player. http://raspi.monaserver.ovh/MonaClients/VideoPlayer/. I also stream the video using RTSP protocol but for it you do not need any server. Also, it is mentioned in documentation that we can use localhost. Is their any need of replay kit in the app to view live streaming?

sherm4n90:
@wasimsafdar If you need it on iOS you can try PLPlayerKit here https://github.com/pili-engineering/PLPlayerKit

shogo4405:
I'm not sure. "RTMP URL parse" is player specific.
For example, http://videojs.com/ is
rtmp://[local-IPAddress]/trainer/live1/**&**streamName

sfaxon:
I've got VLC playback to work with HaishinKit. The problem seems to be with this line:
https://github.com/shogo4405/HaishinKit.swift/blob/978f87901b4e61e26c779bcfac0e12e3f432eeb0/Sources/RTMP/RTMPMuxer.swift#L66-L71

Coming off of the camera `sampleBuffer.decodeTimeStamp` is always `kCMTimeInvalid` so `compositionTime` is always zero. Even if the sampleBuffer had a decodeTimeStamp line 70 is:
`compositionTime = Int32((decodeTimeStamp.seconds - decodeTimeStamp.seconds) * 1000)`
I can't see any way that this value would be non-zero. 

I created a pull request to fix this that is linked above.

PR #451, opened 2018-08-08
setting RTMP composition time to match delta, fixes streaming to VLC.

sfaxon:
This fixes https://github.com/shogo4405/HaishinKit.swift/issues/118


sfaxon:
Can you tell me why this wasn't merged? Is there something I can change that would get it merged in?

I know it's may not follow the spec exactly, but it does fix issues with some players. Does it break any? Could this be a configuration that could be passed in to get this behavior?

PR #478, opened 2018-11-09
Expose new RTMPSocketFactory to allow for custom socket implementation.

gegles:
Default behavior remains unchanged.
Change visibility of the relevant classes/protocols so that custom RTMPSocketCompatible can be provided/implemented from outside of HaiskinKit.


gegles:
Hello @shogo4405. Let me know if exposing this new API is OK with you. It's pretty self-explanatory, but I can provide more context if necessary. I can also tweak and adapt things if needed.

Cheers. G.

gegles:
Any thoughts on this one?

gegles:
That's really a bummer @shogo4405. My thinking is quite different than yours on this. With TCP-replacement protocol such as [SRT](https://github.com/Haivision/srt) or even our own [Aspera FASPStream](https://asperasoft.com/software/streaming/) becoming more and more popular, I suspect this feature would become quite useful to folks out there... 

It's cleanly done via a factory pattern and is even useful currently since there are already 2 types of implementations by default (RTMP and RTMPT). I was especially careful not to affect the normal/default behaviour for the majority of folks who indeed will just use TCP.

If it at all possible, thank you for reconsidering....

Cheers. G.

PR #480, opened 2018-11-13
parse multiple boxes in MP4VisualSampleEntryBox

zhuker:
this will parse multiple child boxes in MP4VisualSampleEntryBox 
sample file here http://kote.videogorillas.com/vmir/GOPR9152.LRV


shogo4405:
👍

PR #482, opened 2018-11-24
Refactor TSWriter

### LINKED ISSUE #450: Crashes when HLS streaming
**Describe the bug**
When streaming HLS, HaishinKit crashes very soon after the start of the streaming.

**To Reproduce**
1. Start macos demo.
2. Select HLS.
3. Start publishing.
4. Open quicktime, hit CMD + L, paste in the URL of the m3u manifest.
5. Wait some time (generally 20 seconds tops)
6. CRASH!

**Expected behavior**
No crash.

**Screenshots**
http://prntscr.com/kg98ox

**Additional context**
It appears that the issue is related to threading. I'll try to look into this some more however I'm not quite certain about the correct use of the threading in the framework.

shogo4405:
1. Fix not compatible to thread-safe (might be fixed #450 )
1. Divide TSWriter -> TSWriter and TSFileWriter

### DIFF
--- a/Sources/Codec/AACEncoder.swift
+++ b/Sources/Codec/AACEncoder.swift
@@ -2,7 +2,7 @@ import AVFoundation
 
 protocol AudioEncoderDelegate: class {
     func didSetFormatDescription(audio formatDescription: CMFormatDescription?)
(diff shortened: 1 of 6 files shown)


vhmth:
@shogo4405 what are the odds of this landing soon? How far off is this from being complete?

Also thank you for tackling this!

PR #495, opened 2019-01-23
fix #366 localRecord is not worked by default

### LINKED ISSUE #366: didFinishWriting not called anymore?
In `v0.8.5+`, it seems like `didFinishWriting(_ recorder: AVMixerRecorder)` of `DefaultAVMixerRecorderDelegate` is no longer called when `RTMPConnection` is closed. While it is in `v0.8.4`.

Anything about that?

wolfcon:
Because property delegate is declared with weak attribute.
```swift
open weak var delegate: AVMixerRecorderDelegate?
```

Solve this Issue #366


wolfcon:
This is useful for recording. Why not merge in. 

In the meantime, there is a warning in Xcode because of it.

PR #506, opened 2019-03-10
drawable orientation to match stream when created

### LINKED ISSUE #505: Camera always as portrait if attached after starting stream
My game lets the user stream only audio or also camera. For this reason, I start the stream with no camera device attached. The app only supports landscape orientation.

When I attach a camera while streaming, it is always in portrait mode, and you have to rotate the device to portrait, then back to landscape to fix this. I have `syncOrientation` enabled always.

I have spent a lot of time on trying to get this to work with `syncOrientation` and setting the orientation only after a delay / only after a camera frame is rendered. Not of these solutions are working, so I must conclude that this behaviour is a bug in the library.

troy-lamerton:
Resolves #505 

The VideoIOComponent.orientation value can be different to the `drawable.orientation` value. This happens when RTMPStream.orientation is set and the drawable is nil.

With this change, after the drawable is created, its orientation is set to match the RTMPStream orientation.

### DIFF
--- a/Sources/Media/VideoIOComponent.swift
+++ b/Sources/Media/VideoIOComponent.swift
@@ -27,7 +27,11 @@ final class VideoIOComponent: IOComponent {
             }
         }
     }
-    var drawable: NetStreamDrawable?
+    var drawable: NetStreamDrawable? = nil {
+        didSet {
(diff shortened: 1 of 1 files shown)


shogo4405:
👍

PR #508, opened 2019-03-12
Typo fix TLS - Transport Layer Security

ankur-revcord:
I think it will be TLS - Transport Layer Security. Not TSL.


shogo4405:
👍

PR #512, opened 2019-03-18
Add the ability to disable the iPhone camera stabilization

psegalen:
There are many reports all over the Internet saying that the internal iPhone camera stabilization system is jamming external ones, for example, it acts against a gimbal stabilization method.
I'd like to use my app with a gimbal such as this one: https://www.dji.com/uk/osmo-mobile-2 but currently the results are shaky because of iOS stabilization behaviour.
Here is a PR to give the possibility to deactivate it within HaishinKit.swift.
Usage:
```
        rtmpStream.captureSettings = [
            "sessionPreset": AVCaptureSession.Preset.hd1280x720.rawValue,
            "continuousAutofocus": true,
            "continuousExposure": true,
            "stabilization": false
        ]
```

### DIFF
--- a/Sources/Media/AVMixer.swift
+++ b/Sources/Media/AVMixer.swift
@@ -35,6 +35,11 @@ public class AVMixer: NSObject {
         get { return videoIO.continuousAutofocus }
         set { videoIO.continuousAutofocus = newValue }
     }
+    
+    @objc var stabilization: Bool {
+        get { return videoIO.stabilization }
+        set { videoIO.stabilization = newValue }
+    }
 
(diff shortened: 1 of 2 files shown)


shogo4405:
👍

psegalen:
Great, thank you for merging it @shogo4405 !
When do you think we'll have a new release having this code?

shogo4405:
@psegalen I tested it. This PR cause a compile error. So I reverted.

スクリーンショット 2019-03-25 1 38 50

psegalen:
Wow that's strange! It was compiling (and running) just fine on my Mac...
Maybe I'm not on the same Swift version, I'll dig this...

PR #516, opened 2019-03-25
Add the ability to disable the iPhone camera stabilization

psegalen:
_Sorry about #512, this one should compile without any error._

There are many reports all over the Internet saying that the internal iPhone camera stabilization system is jamming external ones, for example, it acts against a gimbal stabilization method.
I'd like to use my app with a gimbal such as this one: https://www.dji.com/uk/osmo-mobile-2 but currently the results are shaky because of iOS stabilization behaviour.
Here is a PR to give the possibility to deactivate it within HaishinKit.swift.
Usage:
```
        rtmpStream.captureSettings = [
            "sessionPreset": AVCaptureSession.Preset.hd1280x720.rawValue,
            "continuousAutofocus": true,
            "continuousExposure": true,
            "stabilization": false
        ]

### DIFF
--- a/Sources/Media/AVMixer.swift
+++ b/Sources/Media/AVMixer.swift
@@ -35,6 +35,11 @@ public class AVMixer: NSObject {
         get { return videoIO.continuousAutofocus }
         set { videoIO.continuousAutofocus = newValue }
     }
+    
+    @objc var stabilization: Bool {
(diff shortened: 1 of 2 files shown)


psegalen:
Hi @shogo4405, do you think you could merge this soon please?
I've fixed the error you encountered last time.

PR #518, opened 2019-03-26
fix error when run in xcode10.2

yanyin1986:
Apple released XCode 10.2 this morning.

You will see the error below in file `FLVReader.swift` when you build project with XCode 10.2.
`Mutating method 'readData' may not be used on immutable value 'tag'`


shogo4405:
@yanyin1986 Thank u. so much.

lm2343635:
I have the same problem.

PR #526, opened 2019-04-13
Make RTMPConnection be deallocated on close()

devxoul:
`timer` instance holds the `RTMPConnection` instance so it won't be deallocated after `close()` gets called.


shogo4405:
👍

PR #556, opened 2019-06-25
Add did publish sufficient BW

dawidvdh:
Busy working on a adaptive bitrates, `RTMPStreamDelegate` provides a way to lower the bitrate however I need a way to up the bitrate and I think this solution might work.

Usage:

```swift
let minBitrate: UInt32 = 300 * 1024
let maxBitrate: UInt32 = 2500 * 1024
let incrementBitrate: UInt32 = 512 * 1024

class MyRTMPStreamDelagate: RTMPStreamDelegate {
	@objc var onBWStatusChange: RCTDirectEventBlock?
	
	func didPublishInsufficientBW(_ stream: RTMPStream, withConnection: RTMPConnection) {
		guard var videoBitrate = stream.videoSettings["bitrate"] as? UInt32 else { return }
		print("didPublishInsufficientBW \(videoBitrate)")
		
		videoBitrate = UInt32(videoBitrate / 2)
		if videoBitrate < minBitrate {
			videoBitrate = minBitrate
		}
		stream.videoSettings["bitrate"] = videoBitrate
	}
	
	func didPublishSufficientBW(_ stream: RTMPStream, withConnection: RTMPConnection) {
		guard var videoBitrate = stream.videoSettings["bitrate"] as? UInt32 else { return }
		print("didPublishSufficientBW \(videoBitrate)")
		
		videoBitrate = videoBitrate + incrementBitrate
		if videoBitrate > maxBitrate {
			videoBitrate = maxBitrate
		}
		stream.videoSettings["bitrate"] = videoBitrate
	}
	
	func clear() {
	}
}
```

### DIFF
--- a/Sources/RTMP/RTMPConnection.swift
+++ b/Sources/RTMP/RTMPConnection.swift
@@ -436,6 +436,10 @@ open class RTMPConnection: EventDispatcher {
                 for (_, stream) in streams {
                     stream.delegate?.didPublishInsufficientBW(stream, withConnection: self)
                 }
+            } else if count == 0 {
+                for (_, stream) in streams {
+                  stream.delegate?.didPublishSufficientBW(stream, withConnection: self)
+                }
             }
             previousQueueBytesOut.removeFirst()
         }
--- a/Sources/RTMP/RTMPStreamDelegate.swift
+++ b/Sources/RTMP/RTMPStreamDelegate.swift
@@ -1,4 +1,5 @@
 public protocol RTMPStreamDelegate: class {
     func didPublishInsufficientBW(_ stream: RTMPStream, withConnection: RTMPConnection)
+    func didPublishSufficientBW(_ stream: RTMPStream, withConnection: RTMPConnection)
     func clear()
 }


dawidvdh:
Just noticed that it eats the CPU, gonna try attempt to find an alternative way

mkrn:
Hi Dawid, 
Here's how we've ended up doing it: 
Every second we raise bitrate by 64kbps if it's under default. And once insufficientBW triggered we divide it in 2 like in examples:

```

timer = Timer.scheduledTimer(timeInterval: 1, target: self, selector: #selector(periodicalTasksLaunch), userInfo: nil, repeats: true)

@objc func periodicalTasksLaunch() {
  if let curBitrate = rtmpStream.videoSettings["bitrate"] as? Int {
      bitrate = min(curBitrate + 64 * 1_024, Constants.streamVideoBitrate)
      rtmpStream.videoSettings["bitrate"] = bitrate
    }
}
```

dawidvdh:
@mkrn but dont you feel like having it bundled with `RTMPStreamDelegate` and not having to implement a custom timer would be a better solution? Or am I doing something horribly wrong (learning swift as I go so this is very possible).

shogo4405:
👍

PR #576, opened 2019-08-13
make to Network.framework available

octu0:
I would like to configure RTMPSocket and NetSocket using [Network.framework](https://developer.apple.com/documentation/network).

this pull request Includes **interface changes**
NetSocket used to call back the `listen()` method based on inheritance, but RTMPSocket allows NetSocket to switch between NetSocket and NWSocket(base on Network.framework) -- `NetSocketCompatible` -- so classes that extend NetSocket will be affected.

for example, it needs to be corrected as bellow:

```
// old interface
class Foo: NetSocket {
  :
  override func listen() {
    let data = inputBuffe.toArray(...)
    : // something
  }
}

// new interface
class Bar: NetSocket {
  override init() {
    super.init()
    super.inputHandler = { [weak self] () -> Void in self?.listen() }
  }
  func listen() {
    let data = inputBuffe.toArray(...)
    : // something
  }
}
```

please let me know if there are other good changes.


shogo4405:
👍

PR #581, opened 2019-08-21
Recreate event handlers on connection

leo150:
Right now event handlers are allocated only once in `RTMPSocket.init`, but deallocated every time `NetSocketCompatible.deinitConnection(isDisconnected:)` is called. It leads `RTMPSocket` unable to observe events from socket after first disconnection. 

This PR contains a fix to this issue and reinitializes event handlers on every `connect(withName:port:)` call.


shogo4405:
@leo150 Thank u

PR #652, opened 2020-03-12
[Carthage Support] Turn on 'Enable Testability' for release scheme

innocarpe:
## Context

Hi, our company uses `HaishinKit` via Cocoapod and now try to migrate Carthage.

We made some test stub like below, and need to import `HaishinKit` with `@testable` to access internal properties.

With Carthage an issue occurs when we try to build for testing.

`Module 'HaishinKit' was not compiled for testing`

Screen Shot 2020-03-12 at 6 25 50 PM

## Suggestion

I suggest turning on `ENABLE_TESTABILITY` to `YES` also for the release scheme. 
(Like the screenshot below)
After that, HaishinKit framework which is prebuilt by Carthage (made with Release Scheme) can be used for testing.
I tested it compiled successfully with my forked repo. Also all tests we wrote are passed too.

Screen Shot 2020-03-12 at 6 15 39 PM
 

```swift
@testable import HaishinKit
import SOMETHING

final class RTMPSocketStub: RTMPSocketCompatible {
  ...
  var delegate: RTMPSocketDelegate?
  ...
  var readyState: RTMPSocketReadyState = .uninitialized
  var chunkSizeC: Int = RTMPChunk.defaultSize
  var chunkSizeS: Int = RTMPChunk.defaultSize
  ...
  var securityLevel: StreamSocketSecurityLevel = .none
  var qualityOfService: DispatchQoS = .default

  @discardableResult
  func doOutput(chunk: RTMPChunk, locked: UnsafeMutablePointer<UInt32>?) -> Int {
    Stubber.invoke(doOutput, args: (chunk, locked), default: 0)
  }

  ...
  func connect(withName: String, port: Int) {
    Stubber.invoke(connect, args: (withName, port), default: Void())
  }

  func deinitConnection(isDisconnected: Bool) {
    Stubber.invoke(deinitConnection, args: isDisconnected, default: Void())
  }

  ...
}
```

### DIFF
--- a/HaishinKit.xcodeproj/project.pbxproj
+++ b/HaishinKit.xcodeproj/project.pbxproj
@@ -2261,6 +2261,7 @@
 				DYLIB_COMPATIBILITY_VERSION = 1;
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
+				ENABLE_TESTABILITY = YES;
 				FRAMEWORK_SEARCH_PATHS = (
 					"$(inherited)",
 					"$(PROJECT_DIR)/Carthage/Build/iOS",


innocarpe:
Hm, Apple doesn't recommend to turn on 'Enable Testability' in Release Scheme. It says that option prohibits compiler optimization 🤔

> The Enable Testability build setting should be used only in your Debug configuration, because it prohibits optimizations that depend on not exporting internal symbols from the app or framework. (17732115)

https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW326

innocarpe:
It would be better to add 'open' or 'public' to the protocol or classes for better testability (Stub or Mock) than turning on this option.

shogo4405:
>  Apple doesn't recommend to turn on 'Enable Testability' in Release Scheme.

I think so too. I can’t accept this PR.

I thought in this case 'Test situation' solution is Carthage.private. But I'm not sure how to use it.
https://github.com/Carthage/Carthage/blob/master/Documentation/Artifacts.md#cartfileprivate

PR #658, opened 2020-03-31
Fix SPM support

mfclarke:
Swift Package Manager in Xcode 11 will fail because some import statements are missing. This PR fixes that.


shogo4405:
Thank u.

Issue #682, opened 2020-05-06
[ReasonConfirmed]Reject connection 2 times, then connection succeed.


wolfcon:
### Describe the bug
I am using Wowza to receive rtmp stream and set using authentication with user & password.
When I connected, got ❌
```json
{
    "type": "rtmpStatus",
    "bubbles": false,
    "data": {
        "code": "NetConnection.Connect.Rejected", 
        "description": "[ AccessManager.Reject ] : [ code=403 need auth; authmod=adobe ] : ", 
        "level": "error",
    },
    "target": "HaishinKit.RTMPConnection"
}
```
then got ❌
```json
{
    "type": "rtmpStatus",
    "bubbles": false,
    "data": {
        "level": "error", 
        "code": "NetConnection.Connect.Rejected", 
        "description": "[ AccessManager.Reject ] : [ authmod=adobe ] : ?reason=needauth&user=test&salt=alhJEA==&challenge=Pre+jg==&opaque=Pre+jg=="
    },
    "target": "HaishinKit.RTMPConnection"
}
```
then ✅
**connection succeed.** 😫

### Expected behavior
Connection is normal. Shouldn't get 2 rejection errors. This is led by this kit.  I used `OBS`. All fine no error.

### Reason
```swift
    // rtmpStatusHandler
    ...
    case .some(.connectRejected):
    ...
        switch true {
            case description.contains("reason=nosuchuser"):
                break
            case description.contains("reason=authfailed"):
                break
            case description.contains("reason=needauth"):
                let command: String = RTMPConnection.createSanJoseAuthCommand(uri, description: description)
                connect(command, arguments: arguments)
            case description.contains("authmod=adobe"):
                if user.isEmpty || password.isEmpty {
                    close(isDisconnected: true)
                    break
                }
                let query: String = uri.query ?? ""
                let command: String = uri.absoluteString + (query.isEmpty ? "?" : "&") + "authmod=adobe&user=\(user)"
                connect(command, arguments: arguments)
            default:
                break
            }
```
**Shouldn't throw too much error to user. It does not make sense.**

### Additional context
Already updated HaishinKit to the newest version.

shogo4405:
I'm not sure. That’s how it is with spec.
see https://github.com/shogo4405/HaishinKit.swift/issues/1

wolfcon:
#688 has been refused. 
So. How to solve this issue.

`Shouldn't throw too much error to user. It does not make sense.`

[...]

PR #688, opened 2020-05-15
Use a brand new event/error delegate instead of EventListener

wolfcon:
#682 

Users should not manipulate with `rtmpStatus & ioError`, which are all internal event and too complicated. 
New delegate is more elegant.😁


shogo4405:
> New delegate is more elegant.😁

I don't think so.

wolfcon:
How about block? Just pack `Event` to `State` without complicated register/unregister

PR #675, opened 2020-04-27
Add mirrored orientation adjustment for front facing camera when using MTHKView.

hambly:
This addresses https://github.com/shogo4405/HaishinKit.swift/issues/674

This PR matches the functionality in GLHKView. When rendering the video preview, check if the front facing camera is being used, and if so, apply the "up mirrored" orientation.

From GLHKView:
```
if position == .front {
  currentStream?.mixer.videoIO.context?.draw(displayImage.oriented(forExifOrientation: 2), in: inRect, from: fromRect)
} else {
  currentStream?.mixer.videoIO.context?.draw(displayImage, in: inRect, from: fromRect)
}
```

Suggested change for consistency in MTHKView:
```
let adjustedImage = position == .front ? image.oriented(forExifOrientation: 2) : image
let scaledImage: CIImage = adjustedImage
  .transformed(by: CGAffineTransform(translationX: translationX, y: translationY))
  .transformed(by: CGAffineTransform(scaleX: scaleX, y: scaleY))
```

PR #704, opened 2020-06-18
Feature: add property `isMirrored` in MTHKView and GLHKView

allenlinli:
# Description
I add `isMirrored` property for GLHKView and MTHKView, so the view can be mirrored when `rtmpStream.captureSettings[.isVideoMirrored]` is set true.

I did not add a mirror button on the UI, because there's no button for setting `rtmpStream.captureSettings[.isVideoMirrored]` to true yet. We may need manual testing for this part, or I can add the UI later if needed.

I hope this new property is clear and useful enough.

### DIFF
--- a/Platforms/iOS/MTHKView.swift
+++ b/Platforms/iOS/MTHKView.swift
@@ -13,6 +13,7 @@ open class MTHKView: MTKView, NetStreamRenderer {
 
     var position: AVCaptureDevice.Position = .back
     var orientation: AVCaptureVideoOrientation = .portrait
+    var isMirrored: Bool = false
 
     var displayImage: CIImage?
     weak var currentStream: NetStream? {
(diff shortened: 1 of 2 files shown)


allenlinli:
This PR is the same as https://github.com/shogo4405/HaishinKit.swift/pull/701 
That PR was closed by mistake.

Furthermore, I made the properties open.
I am not sure either `open` or `public` is more suitable for this property.

Platforms/iOS/MTHKView.swift
```diff
@@ -103,9 +104,18 @@ extension MTHKView: MTKViewDelegate {
...
             .transformed(by: CGAffineTransform(scaleX: scaleX, y: scaleY))
+
+        if isMirrored {
+            if #available(iOS 11.0, *) {
+                scaledImage = scaledImage.oriented(.upMirrored)
+            } else {
+                scaledImage = scaledImage.oriented(forExifOrientation: 2)
+            }
+        }
+
```
hambly, line 118: The orientation needs to happen before the transformation. @shogo4405

allenlinli: Hi @hambly, I am not familiar with the CGAffineTransform part, can you explain more about why does the orientation needs to happen before the transformation?
The appended PR and comments are welcome, too.

hambly: Orientation is inherent to the media, the transformation is part of rendering for display, in particular the translation aspect.

allenlinli: So you mean the results would be the same here, but we should usually do orientation before transformation for the process, correct?

hambly: The results are not the same here, this PR did not correctly configure the mirrored front-facing camera, as the translation becomes incorrect upon mirroring.

See https://github.com/shogo4405/HaishinKit.swift/pull/675

allenlinli: I am not sure how can the translation becomes incorrect upon mirroring...
May I ask how did the PR https://github.com/shogo4405/HaishinKit.swift/pull/675/files being closed? Is it a correct PR that should be merged?

You can propose a PR to amend this one, or I will clarify it with and give a PR based on your suggestion.

Issue #606, opened 2019-10-14
Crash on NetSocket.swift


Goule:
Hello, 

**Describe the bug**
Crashlytics is reporting a crash on NetSocket.swift line 103 (`NetSocket.doOutputProcess(_:maxLength:)`). 

**To Reproduce**
I can't reproduce it but it occured 10 times in 1 week in production (not many users).

**Smartphone (please complete the following information):**
Exemple of impacted devices : 
-  12.4.1 (16G102) / iPhone 7
- 12.3.1 (16F203) / iPhone 6s
- 12.4.2 (16G114) / iPad Air
- 11.4.1 (15G77) / iPhone 6

**Additional context**
Here is the logs from Crashlytics : 

```
Crashed: com.haishinkit.HaishinKit.NetSocket.output
0  CoreFoundation                 0x1d8cdd3e0 CFHash + 372
1  CoreFoundation                 0x1d8d70780 CFBasicHashGetCountOfKey + 204
2  CoreFoundation                 0x1d8cde8bc CFSetContainsValue + 116
3  CoreFoundation                 0x1d8cd6e18 CFRunLoopRemoveSource + 164
4  CFNetwork                      0x1d93ebc98 SocketStream::write(__CFWriteStream*, unsigned char const*, long, CFStreamError*) + 592
5  CoreFoundation                 0x1d8cec0c0 CFWriteStreamWrite + 300
6  HaishinKit                     0x10138dba4 NetSocket.doOutputProcess(_:maxLength:) + 103 (NetSocket.swift:103)
7  HaishinKit                     0x10138d98c closure #1 in NetSocket.doOutput(data:locked:) + 49 (NetSocket.swift:49)
8  HaishinKit                     0x1013355d8 thunk for @escaping @callee_guaranteed () -> () (<compiler-generated>)
9  libdispatch.dylib              0x1d8788a38 _dispatch_call_block_and_release + 24
10 libdispatch.dylib              0x1d87897d4 _dispatch_client_callout + 16
11 libdispatch.dylib              0x1d8732320 _dispatch_lane_serial_drain$VARIANT$mp + 592
12 libdispatch.dylib              0x1d8732e3c _dispatch_lane_invoke$VARIANT$mp + 428
13 libdispatch.dylib              0x1d873b4a8 _dispatch_workloop_worker_thread + 596
14 libsystem_pthread.dylib        0x1d8969114 _pthread_wqthread + 304
[...]

PR #733, opened 2020-08-12
Might be fixed crash on NetStream and decrease memory usages.

shogo4405:
#606 Might be fixed crash on NetStream.


Sources/Net/NetSocket.swift
```diff
@@ -186,7 +186,7 @@ open class NetSocket: NSObject {
     }
 
     func close(isDisconnected: Bool) {
-        outputQueue.async {
+        inputQueue.async {
```
skahack, line 189: I think the loop has to be stopped first.

Issue #759, opened 2020-09-27
Audio and video becoming out of sync when playing


casper6479:
**Describe the bug**
When using play feature from HaishinKit, audio and video becomes out of sync, when streamer have connection paused and resumed.

**To Reproduce**
1. You can start publishing from HaishinKit or any other stream source.
2. Start playing stream with HaishinKit.
3. Pause(or stop) publishing for a couple of seconds, server must not send connection close event, then resume publishing. Or if you are publishing with HaishinKit, just move app with camera (from publisher) into background, then open it back.
4. Player client will have audio or video delay (i think sometimes video, sometimes audio). Other players works just fine in this case. Also if you have choppy playback for some reason, you will have the same issue.

**Expected behavior**
Video frames (or audio frames) should drop in some cases, but looks like they are not. Looks like video-decoding process starting and waiting until first frame will be decoded and playing it, but audio could be already decoded and running without any sync with video.

**Smartphone (please complete the following information):**
 - Device: iPhone 11 Pro, iPhone 7
 - OS: iOS 13, iOS 14

**Additional context**
Error can be reproduced in demo app.

mikolajadamowicz:
Yup, we have exactly the same issue both on iOS 13 and 14 across all the devices.

casper6479:
I've fixed it in some cases, if i reinit rtmpconnection on streamBegin event from RTMPUserControlMessage. But that's definitely bad and can't fix so much cases

casper6479:
My solution still breaks sync for some cases in normal viewing, but 100% it helps when streamer pausing and unpausing stream. @shogo4405 really need your help

PR #761, opened 2020-10-02
Use DisplayLink timestamp instead of clockReference

casper6479:
Using clockReference duration may break time logic and audio video sync if video was freezed on streamer side


casper6479:
Fixes #759 at least for my cases. You may need to change clockReference logic, but honesty, im not sure you really needed to use clockReference

casper6479:
It fixes issues when streamer pausing camera, but in normal playing, after some time, video could be out of sync again. So we have some logical problem here

PR #878, opened 2021-06-16
Updated Stream sponsorship URL

nparsons08:
@shogo4405 small update to Stream sponsorship URL


shogo4405:
@nparsons08 OK. Thanks.

PR #884, opened 2021-07-03
Error in Xcode 12.5 Canvas (SwiftUI)

gevorgyanvahagn:
It shows an error in SwiftUI canvas. This change fixes the error.
`overlapping accesses to 'self.data', but modification requires exclusive access; consider copying to a local variable`


shogo4405:
Well, It seemes that contains performance issue. I can't accept.
How can I reproduce it?

Issue #896, opened 2021-08-11
Video Toolbox Framework Requires tvOS 10.2 or higher


shamanskyh:
**Describe the bug**
The Swift Package specifies a minimum version of tvOS 10.0. However, HaishinKit depends on Apple's [Video Toolbox framework](https://developer.apple.com/documentation/videotoolbox) which was not made available until tvOS 10.2.

**To Reproduce**
1. Create a tvOS application and add HaishinKit as a Swift Package dependency.

**Expected behavior**
The build completes as expected

**Actual behavior**
The build fails because symbols from the Video Toolbox framework are unavailable in tvOS versions earlier than tvOS 10.2. Even though my application has a tvOS 14 minimum target, HaishinKit fails to build as a dependency because of its Package.swift.

**Potential Fixes**
* Make the minimum tvOS version 10.2
* Guard the functionality that relies on the Video Toolbox framework

PR #897, opened 2021-08-11
Update Package.swift to require tvOS 10.2 or higher

shamanskyh:
Update the Swift package file to correctly specify the minimum version in MARKDOWN.md. Solves #896


shamanskyh:
Actually, this has a build error. Will reopen with a fixed version -- apologies.

PR #903, opened 2021-09-07
fix: render view x origin when using isMirrored option with MTHKView

MarcFerry:
When using MTHKView with _resizeAspectFill_ videoGravity, the x origin is set incorrectly with the _isMirrored_ option (tested on iPad Pro landscape).
Applying the mirror effect before resizing the image fixes the issue.

Screen Shot 2021-09-07 at 16 52 51


shogo4405:
Thank you.

Issue #888, opened 2021-07-15
Error in SwiftUI Canvas


floriangbh:
Hello, 

**Describe the bug**

HaishinKit doesn't permit to run the canvas (with SwiftUI) in Xcode. 

Here is the error when project is compiled for the canvas : 

<img width="576" alt="Capture d’écran 2021-07-15 à 11 01 01" src="https://user-images.githubusercontent.com/14054997/125760964-73963802-b1bf-477f-af47-66fffa363afd.png">

<img width="1001" alt="Capture d’écran 2021-07-15 à 11 08 25" src="https://user-images.githubusercontent.com/14054997/125761895-9a4bb212-14ef-41c2-8a16-73e6d9a00e62.png">

**To Reproduce**
1. Open HaishinKit
2. Put iOS 13+ target 
2. Create a new "SwiftUI View" file
3. "Resume" the view in the canvas 
4. See an error in the diagnostics button

**Expected behavior**
Should compile and permit to work with SwiftUI view.

**Desktop (please complete the following information):**
 - Xcode: 12.5
 - MacOS 11.4
 
**Additional context**

This PRs: https://github.com/shogo4405/HaishinKit.swift/pull/884 should resolve this issue.

PR #906, opened 2021-09-13
Fix warnings (preventing any SwiftUI 'preview' views showing in xcode)

scornflake:
Thought I'd do a bit of tidy up. Only checked against latest Swift compiler though... useful?


Sources/Net/NetSocket.CircularBuffer.swift
```diff
@@ -45,7 +45,8 @@ extension NetSocket {
                 self.locked = locked
             }
             let length = min(count, capacity - tail)
-            return self.data.withUnsafeMutableBytes { (bytes: UnsafeMutableRawBufferPointer) -> Bool in
```
shogo4405, line 48: I have a concern about performance issue.
I will change a struct to class.

shogo4405: So I will reject this issue.

scornflake: I believe it would be the same, isn't it a copy on write? managed by the compiler?  So I think the previous (current) version was the same (still a copy on write), just unsafe. Here; it's expliit.   But: I havn't profiled this.

Goule:
Looks like the same issue as : https://github.com/shogo4405/HaishinKit.swift/issues/888

@scornflake did you find a workaround for SwiftUI ?

PR #913, opened 2021-09-26
fixes compilation through SPM

AndrewSB:
as i tried to compile the project as a dependency through SPM, i was seeing issues due to CADisplayLink not being found, and nstry being undefined.

i fixed the CADisplayLink issue, i think for nstry we need to add a entry to the Package.swift that indicates where the .h project headers can be found


shogo4405:
Thank you. I can't remove `nstry`. I think another way.

PR #951, opened 2022-01-07
Feature/swiftui porting

y2kpaulh:
Hello. I am a user of HaishinKit library. I don't know if this source will be of any help, but I'll give you an example source for working with SwiftUI target that I've been working on.

Thanks for your hard work.

### DIFF
--- a/Examples/iOSSwiftUI/ContentView.swift
+++ b/Examples/iOSSwiftUI/ContentView.swift
@@ -1,10 +1,36 @@
 import HaishinKit
+import AVFoundation
+import VideoToolbox
 import SwiftUI
 
 struct ContentView: View {
+    @ObservedObject var viewModel = ViewModel()
+    
+    private var lfView: MTHKSwiftUiView!
+    private var menuView: MenuView!
+    
+    init() {
+        viewModel.config()
+        lfView = MTHKSwiftUiView(rtmpStream: $viewModel.rtmpStream)
+        menuView = MenuView(viewModel: viewModel)
+    }
+    
     var body: some View {
-        Text("Hello, world!!")
-            .padding()
+        ZStack {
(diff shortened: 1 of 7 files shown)


shogo4405:
Thank you for this PR. Great!

Examples/iOSSwiftUI/ContentView.swift
```diff
@@ -1,10 +1,36 @@
 import HaishinKit
+import AVFoundation
+import VideoToolbox
 import SwiftUI
 
 struct ContentView: View {
+    @ObservedObject var viewModel = ViewModel()
```
Goule, line 7: Shouldn't that be a StateObject instead?

Examples/iOSSwiftUI/ContentView.swift
```diff
@@ -1,10 +1,36 @@
...
+    
+    private var lfView: MTHKSwiftUiView!
+    private var menuView: MenuView!
+    
+    init() {
+        viewModel.config()
+        lfView = MTHKSwiftUiView(rtmpStream: $viewModel.rtmpStream)
+        menuView = MenuView(viewModel: viewModel)
```
Goule, line 15: The problem with doing this is that the init can pass multiple times, especially when using native navigation with SwiftUI. Struct can be instantiate multiple time

Goule:
Hello @y2kpaulh, 

Do you have a more recent example ? 
I tried to implement it with more recent and complex project. 
With the new NavigationPath navigation, the init pass multiple time and nothing work. 

I think there are some reasons : 
- the ObservableObject should be an StateObject 
- the code in the Init, should not be here, SwiftUI can crash et re create the struct many time with the navigation for example 
But I can't figure out how to make it work properly..

PR #967, opened 2022-02-18
Fix: encodeSampleBuffer do not render what is encoded

chatelgu:
Hi, 

Playing with Effects, I found out that some effects was not rendered on mobile screen, but apply to the streamed video.

I propose this change use the same `imageBuffer ??  buffer` in the line 419 of [Sources/Media/AVVideoIOUnit.swift](https://github.com/shogo4405/HaishinKit.swift/compare/main...chatelgu:fix-preview-if-not-in-place?expand=1#diff-b8cbb74787769dd699b013d9eb733d871c9eb84c0bf818f0491d986e8eb5dcd4) than for the `encoder` and `recorder`.

By the way, I didn't get well why imageBuffer is not used if the effects does not change the size of the image.

### DIFF
--- a/Platforms/iOS/HKView.swift
+++ b/Platforms/iOS/HKView.swift
@@ -46,7 +46,7 @@ open class HKView: UIView {
         }
     }
     var position: AVCaptureDevice.Position = .front
-    var currentSampleBuffer: CMSampleBuffer?
+    var currentImageBuffer: CVImageBuffer?
 
     private weak var currentStream: NetStream? {
         didSet {
@@ -96,7 +96,7 @@ open class HKView: UIView {
 
 extension HKView: NetStreamRenderer {
     // MARK: NetStreamRenderer
-    func enqueue(_ sampleBuffer: CMSampleBuffer?) {
+    func enqueue(_ sampleBuffer: CVImageBuffer?) {
     }
 }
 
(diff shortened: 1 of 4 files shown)


shogo4405:
I won't change CMSampleBuffer -> CVImageBuffer.

> Playing with Effects, I found out that some effects was not rendered on mobile screen, but apply to the streamed video.

How can I reproduce it. I want to see exsample code.

chatelgu:
Please try to crop an image in an effect : https://developer.apple.com/documentation/coreimage/ciimage/1437833-cropped

chatelgu:
To be more precise to reproduce my issue. I give you that sample (dummy) code:

```swift
final class CropEffect: VideoEffect {
    override func execute(_ image: CIImage, info: CMSampleBuffer?) -> CIImage {
        let rect = image.extent
        return image.cropped(to: CGRect(x: 0, y: 0, width: rect.width/2, height: rect.height/2))
    }
}
```

and then `if rtmpStream.registerVideoEffect(CropEffect()) { print("register effect") }` should reproduce the issue I have.

PS: Sorry for the late answer.

shogo4405:
I understand this issue. But I need CMSampleBuffer. 
I can't accept this pr. Thank you.

PR #971, opened 2022-03-17
Ftr improve stream quality ios

HamzaKiyani43:
i have changed encoder height and width and drawHiererchy function so it will take whole mobile screen and better quality to go live.


shogo4405:
This pr contains magick number.
 - CGRect(x: 73)
 - encoder.width = Int32(1920)

I can't accept this commit.

PR #1041, opened 2022-08-01
Remove stop recording on stop publishing

Goule:
Hello, 

Since this PR https://github.com/shogo4405/HaishinKit.swift/pull/1035 local recording is not really related with publishing. 

This PR let the user decide when the recording should be stopped. 

Goal : allow record even on a network cut


shogo4405:
Thank you.

Issue #892, opened 2021-07-28
Metadata don't send if you append  as CMSampleBuffer 


andrew3213221:
**Describe the bug**
RTMP stream
Metadata don't send if you append  as CMSampleBuffer 

**To Reproduce**
Steps to reproduce the behaviour:
connect to RTMP server and use input as sample buffer 
`RTMPStream.appendSampleBuffer(CMSampleBuffer, AVMediaType)`
the problem is when message with data send 
use function createMetaData which check if input device was used 
but appendSampleBuffer doesn't set input device as function attachCamera

**Expected behavior**
When you use own capture session with output metadata must be included also

**Desktop (please complete the following information):**
 - OS: iOS
 - Version 1.1.6

shogo4405:
I have no good idea.  Please override `RTMPStream`.
```
class YourRTMPStream: RTMPStream {
    override func createMetadata() -> ASObject  {
           ...
    }
}
```
https://github.com/shogo4405/HaishinKit.swift/pull/1074

Close as won't fix.

PR #1074, opened 2022-10-13
Fix metadata not coming through when sending sampleBuffers

charmingtoad:
## Type of change

### DIFF
--- a/Sources/RTMP/RTMPStream.swift
+++ b/Sources/RTMP/RTMPStream.swift
@@ -430,14 +430,17 @@ open class RTMPStream: NetStream {
     open func createMetaData() -> ASObject {
         metadata.removeAll()
         #if os(iOS) || os(macOS)
+        if let _: AVCaptureInput = mixer.videoIO.input {
             metadata["width"] = mixer.videoIO.codec.width
             metadata["height"] = mixer.videoIO.codec.height
             metadata["framerate"] = mixer.videoIO.fps
             metadata["videocodecid"] = FLVVideoCodec.avc.rawValue
             metadata["videodatarate"] = mixer.videoIO.codec.bitrate / 1000
-        
+        }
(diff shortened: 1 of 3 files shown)


charmingtoad:
I had the same issue as https://github.com/shogo4405/HaishinKit.swift/issues/892 and fixed it with this.

Sources/RTMP/RTMPStream.swift
```diff
@@ -430,17 +430,14 @@ open class RTMPStream: NetStream {
     open func createMetaData() -> ASObject {
         metadata.removeAll()
         #if os(iOS) || os(macOS)
-        if let _: AVCaptureInput = mixer.videoIO.input {
```
shogo4405, line 433: This is not good. This code will send video and audio metadata always. I don't want send.

shogo4405 (changes requested):

Sources/RTMP/RTMPStream.swift
shogo4405: Please override your code.
```
class YourRTMPStream: RTMPStream {
    override func createMetadata() -> ASObject  {
           ...
    }
}
```

charmingtoad: I see, I didn't think of the case that a user might want to only send video or audio but not both.
I made a new class where user can set flags to specify what kind of sampleBuffer they intend to send.
Hope this is okay, let me know if more changes are needed. Thanks!

charmingtoad: Ah, I have the issue that I can't use this in my application code because mixer.videoIO and mixer.audioIO have 'internal' protection level, so my app cannot access them. Is there any way to send metadata for apps that use the appendSampleBuffer method?

charmingtoad:
ああ、これで分かった。教えてくれてどうもありがとうございます。このPRを閉めます。

PR #1079, opened 2022-11-03
Fix: session is reinitialized every frame

leo150:
## Description & motivation

`VideoCodec` never sets `invalidateSession` to `false`. As result, `session` is reinitialized after every `inputBuffer` call.

Some misc updates:
* Add AppCode folder to .gitignore
* Check status after setting video codec options and call appropriate delegate method

## Type of change


shogo4405:
Exactly. Thank you.

PR #1133, opened 2023-02-01
Feature: Support for custom recorder implementation

peterpaulis:
## Description & motivation

This adds support to easily extend and implement a custom recorder, following the IORecorderHandler protocol... such a custom recorder can choose to record different outputs to different files, having custom logic how the recording should be done...

## Type of change
- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/Sources/Media/IOAudioUnit.swift
+++ b/Sources/Media/IOAudioUnit.swift
@@ -74,7 +74,6 @@ final class IOAudioUnit: NSObject, IOUnit {
         guard let sampleBuffer = sampleBuffer.muted(muted) else {
             return
         }
-        mixer?.recorder.appendSampleBuffer(sampleBuffer, mediaType: .audio)
         codec.encodeSampleBuffer(sampleBuffer)
     }
 }
@@ -120,6 +119,7 @@ extension IOAudioUnit: AVCaptureAudioDataOutputSampleBufferDelegate {
         guard mixer?.useSampleBuffer(sampleBuffer: sampleBuffer, mediaType: AVMediaType.audio) == true else {
             return
         }
+        mixer?.recorder.handleAudioOutput(output, didOutput: sampleBuffer, from: connection)
         appendSampleBuffer(sampleBuffer)
     }
 }
--- a/Sources/Media/IOMixer.swift
+++ b/Sources/Media/IOMixer.swift
@@ -71,8 +71,8 @@ public class IOMixer {
     #endif
     public private(set) var isRunning: Atomic<Bool> = .init(false)
     /// The recorder instance.
-    public lazy var recorder = IORecorder()
-
+    public lazy var recorder: IORecorderHandler = IORecorder()
+    
     /// Specifies the drawable object.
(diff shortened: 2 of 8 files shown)


shogo4405:
It is unacceptable because I cannot be maintained in the future. 
Even if you design it like this, in most cases the design will be complicated and no one will use it.

peterpaulis:
It adds abstraction to the recorder, which is the main point somebody wants to customize the library... the addition of more inputs, like screensharing, multicamp, mixing, provides the need for the recorder to be customized by the programmer

also the library is more usable in context of other libraries, as the capture output can just be linked to the previous library

like bellow is an implementation which uses an already existing recorder and the connection would be impossible without the recorder abstraction

to sum up the benefits, the programmer is now open to
```
- use a custom dash recorder
- separately record each input
- record unchanged input
- use in conjunction with other recording libraries
- apply other processing
```

without the recorder abstraction, the recorder is not expandable

```
@objc
class ForwardIORecorder: NSObject, IORecorderHandler {
    
    weak var coreRecorder: M6DashRecorder?
    
    init(coreRecorder: M6DashRecorder?) {
        self.coreRecorder = coreRecorder
    }
    
    func prepareForSettings(_ settings: [AVMediaType : [String : Any]]) {
        
        
        
    }
    
    func handleCameraOutput(_ output: AVCaptureOutput, didOutput sampleBuffer: CMSampleBuffer, from connection: AVCaptureConnection) {
        coreRecorder?.externalCaptureOutput(output, didOutputSampleBuffer: sampleBuffer, from: connection, isVideo: true)
    }
    
    func handleMultiCameraOutput(_ output: AVCaptureOutput, didOutput sampleBuffer: CMSampleBuffer, from connection: AVCaptureConnection) {
        
    }
    
    func handleAudioOutput(_ output: AVCaptureOutput, didOutput sampleBuffer: CMSampleBuffer, from connection: AVCaptureConnection) {
        coreRecorder?.externalCaptureOutput(output, didOutputSampleBuffer: sampleBuffer, from: connection, isVideo: false)
    }
    
    func handleScreenOutput(_ sampleBuffer: CMSampleBuffer) {
        
    }
    
    func handleMixedVideoOutput(_ pixelBuffer: CVPixelBuffer, withPresentationTime: CMTime, sampleBuffer: CMSampleBuffer, multicamSampleBuffer: CMSampleBuffer?) {
        
    }
    
    func handleStreamOutput(_ sampleBuffer: CMSampleBuffer, withType: AVMediaType) {
        
    }
    
    public private(set) var isRunning: HaishinKit.Atomic<Bool> = .init(false)

    @objc
    func startRunning() {
        guard !self.isRunning.value else {
            return
        }
        self.isRunning.mutate { $0 = true }
    }
    
    @objc
    func stopRunning() {
        guard self.isRunning.value else {
            return
        }
        self.isRunning.mutate { $0 = false }
    }
    
    
}
```

peterpaulis:
and to be honest i don't see the point why it "cannot be maintained"... the recorder is basically a separate part of the lib, and this change supports that

PR #1153, opened 2023-03-14
Added handling sample of Adaptive bitrate streaming.

shogo4405:
## Description & motivation
Added handling sample of Adaptive bitrate streaming. 

## Type of change


sangriel:
even if [stream.videoSettings.bitRate = max(videoBitRate, 64 * 1000)] has been changed does it really have affect on socket.outputBuffer size? i applied your code but it seems there is no effect on actual bytes transferring to rtmp server
And outputBuffer is continually queued up without being released.

Issue #1210, opened 2023-05-26
[RTMPNWSocket] The video bandwidth is insufficient upon reconnection.


shogo4405:
### Describe the bug

https://github.com/shogo4405/HaishinKit.swift/issues/1203#issuecomment-1564472679

An issue I have is after reconnection the way publishSufficientBWOccured is quite different:
It's getting sufficient but currentBytesOut are consistently lower,
For example, setting bitrate to 2765000 (2.7 mbps) consistently pushing out ~501216 bits per second (500kbps)
next second it still pushes only 500kbps or so even with higher bitrate

### To Reproduce

In Sample app (SwiftUI iOS)
Start with Wifi + 5G
Swipe to controls, turn off WiFi
I'm getting these log messages:

### Expected behavior

N/A

### Version

N/A

mkrn:
I tested it some more, and it definitely is hard to impossible to re-establish connection while on LTE/5G now.. Or the connection is much slower once re-established, even if initially it was good speed (same location). 
It seems to re-connect well with good speed on WiFi.

iPhone 13 Pro
Latest HaishinKit (main)
iOS 16.1
Vodafone Portugal

mkrn:
Ok, here's the log where I have the reconnection stuck forever, but I can't yet reproduce it in your Example app.. So maybe it's something very rare, but can anything be done with it to timeout the reconnection?

```

2023-06-06 19:40:06.702179+0100 appnew[14115:2955040] Target bitrate 2764800
2023-06-06 19:40:07.593487+0100 appnew[14115:2955995] [] nw_flow_prepare_output_frames Failing the write requests [57: Socket is not connected]
2023-06-06 19:40:07.593587+0100 appnew[14115:2955995] [connection] nw_write_request_report [C25] Send failed with error "Socket is not connected"
2023-06-06 19:40:07.594403+0100 appnew[14115:2955995] [] nw_flow_prepare_output_frames Failing the write requests [57: Socket is not connected]
2023-06-06 19:40:07.594486+0100 appnew[14115:2955995] [connection] nw_write_request_report [C25] Send failed with error "Socket is not connected"
2023-06-06 19:40:07.594801+0100 appnew[14115:2955995] [] nw_flow_prepare_output_frames Failing the write requests [57: Socket is not connected]
2023-06-06 19:40:07.594855+0100 appnew[14115:2955995] [connection] nw_write_request_report [C25] Send failed with error "Socket is not connected"
[...]

PR #1221, opened 2023-06-23
Fixed hanging connections. Added viability checker to NWSocket.

mkrn:
- Fixed a situation when socket is hanging while trying to close a non-viable connection
- Fixed a situation when the connection is closed but RTMPConnection is still trying to send data.
 

## Description & motivation
Fixes https://github.com/shogo4405/HaishinKit.swift/issues/1210 

## Type of change


Sources/RTMP/RTMPConnection.swift
```diff
@@ -456,6 +458,19 @@ open class RTMPConnection: EventDispatcher {
...
+            zeroBytesOutCounter += 1
+            // If currentBytesOutPerSecond has been 0 for the last 5 seconds
+            if zeroBytesOutCounter >= 5 {
+                print("STALE connection, force close")
+                self.close(isDisconnected: true) // call close function
+                zeroBytesOutCounter = 0 // reset counter
+            }
+        } else {
+            zeroBytesOutCounter = 0 // reset counter
+        }
```
shogo4405, line 472: I believe it is abnormal for there to be no output for 5 seconds during publishing. On the other hand,  Playback feature it is a possible scenario. If I accept it as it is, it will be considered a bug.

shogo4405: It seems like a good idea to implement the PING/PONG mechanism for RTMP.

Sources/RTMP/RTMPNWSocket.swift
```diff
@@ -103,16 +110,7 @@ final class RTMPNWSocket: RTMPSocketCompatible {
...
-        if connection.state == .ready {
-            outputQueue.async {
-                let completion: NWConnection.SendCompletion = .contentProcessed { (_: Error?) in
-                    self.connection = nil
-                }
-                connection.send(content: nil, contentContext: .finalMessage, isComplete: true, completion: completion)
-            }
-        } else {
-            self.connection = nil
-        }
```
shogo4405, line 115: This process was created specifically to address this issue, so simply removing it would cause the problem to reoccur.

shogo4405 (changes requested):
I think it would be a good idea to introduce connection?.viabilityUpdateHandler, so I will try.

shogo4405:
Thank you for providing the PR. I will refer to it and try to make the necessary fixes on my end as well.

Sources/RTMP/RTMPConnection.swift
mkrn: Yes  maybe ping pong is a better idea, I couldn't tell if it's standard for RTMP.
For me the stale connections happen with this output:

2023-06-22 16:10:31.674537+0100 appnew[1163:121214] [tcp] tcp_input [C15:1] flags=[R] seq=1983586932, ack=0, win=0 state=CLOSE_WAIT rcv_nxt=1983586933, snd_una=1331617881


CLOSE_WAIT means that the local end of the connection has received a FIN from the other end, but the OS is waiting for the program at the local end to actually close its connection. The problem is your program running on the local machine is not closing the socket.

After CLOSE_WAIT notice from NWConnection provided by OS, RTMP Connection is still active and sends publishSufficientBWOccured,  but currentBytesOutPerSecond are 0  and it can last for long time  1+ minutes.

I don't know where CLOSE_WAIT is coming from.. maybe server issue. Tried with different servers,  RTMP, RTMPS, with balancer, and without, but still getting CLOSE_WAIT when reconnecting quickly.

Sources/RTMP/RTMPNWSocket.swift
mkrn: I see... What was happening is: connection was stale so end message could not be sent..  so it was stuck forever,  completion handler was never reached.

Maybe with viabilityUpdateHandler check for viability this could work

Examples/iOS/Preference.swift
```diff
@@ -1,6 +1,6 @@
 struct Preference {
     static var defaultInstance = Preference()
 
-    var uri: String? = "rtmp://192.168.1.6/live"
-    var streamName: String? = "live"
+    var uri: String? = "rtmps://tempbalancer.eventlive.pro:443/live"
```
leo150, line 4: There are some changes in this commit that are not related to the fix itself. Could you please remove it? This url, new DEVELOPMENT_TEAM, changing maxRetryCount, commenting code, new padding in stateDidChange. Thank you

shogo4405:
I have incorporated the elements of this PR into the main branch. The only remaining task is to properly implement RTMP PING/PONG. You can go ahead and close the PR. Thank you very much.

PR #1232, opened 2023-07-11
Invoke delegate the stream(_ stream: NetStream, didOutput video: CMSa…

shogo4405:
…mpleBuffer) method.

## Type of change

### DIFF
--- a/Examples/iOS/PlaybackViewController.swift
+++ b/Examples/iOS/PlaybackViewController.swift
@@ -16,6 +16,7 @@ final class PlaybackViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         rtmpStream = RTMPStream(connection: rtmpConnection)
+        rtmpStream.delegate = self
     }
 
     override func viewWillAppear(_ animated: Bool) {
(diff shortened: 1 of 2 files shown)


zaliva08:
After the changes you made, for me still do not work out the delegate and does not come CMSampleBuffer.
    override func viewDidLoad() {
        super.viewDidLoad()
        rtmpConnection.delegate = self
        rtmpStream.delegate = self
    }
extension MAnchorVC: NetStreamDelegate {
    func stream(_ stream: HaishinKit.NetStream, didOutput video: CMSampleBuffer) {
        guard !self.isStopped else { return }
        XMagicManager.share().processData(withCpuFuc: video)
    }
    
    How else can I get CMSampleBuffer ?

zaliva08:
Add please delegate?.mixer(self, didOutput: sampleBuffer)
  You should do this in func useSampleBuffer
    
    func useSampleBuffer(sampleBuffer: CMSampleBuffer, mediaType: AVMediaType) -> Bool {
        switch mediaSync {
        case .video:
            if mediaType == .audio {
                return !videoTimeStamp.seconds.isZero && videoTimeStamp.seconds <= sampleBuffer.presentationTimeStamp.seconds
            }
            if videoTimeStamp == CMTime.zero {
                videoTimeStamp = sampleBuffer.presentationTimeStamp
            }
            delegate?.mixer(self, didOutput: sampleBuffer)
            return true
        default:
            return true
        }
    }

shogo4405:
This feature is primarily designed for the behavior during playback and becomes part of the specification.

zaliva08:
I had to implement a library of effects. I needed SampleBuffer for this from rtmpStream. By adding this delegate in library I was able to get the result I need. 

But I added the library HashimKit as static because of this. It would be nice if you would correct yourself then I will go back to use Cartfile

PR #1260, opened 2023-08-07
Add a stale connection detection. Fix falsy publishSufficientBWOccured

### LINKED ISSUE #1210: [RTMPNWSocket] The video bandwidth is insufficient upon reconnection.

mkrn:
## Description & motivation

Fixes #1210 
For some reason after forced disconnect from WiFi and switch to Cellular the connection would hang.
The fix is to detect when there are 5s of 0 bytes sent in a row. When this happens, we force reconnect. 
Also, there's a bug that sends publishSufficientBWOccured event in a scenario where there were many 0 bytes sent periods in a row. This problem caused some adaptive bitrates to increase bitrate when connection was stale.

## Type of change


Sources/RTMP/RTMPConnection.swift
```diff
@@ -455,12 +457,26 @@ open class RTMPConnection: EventDispatcher {
...
+            zeroBytesOutCounter += 1
+            // If currentBytesOutPerSecond has been 0 for the last 5 seconds
+            if zeroBytesOutCounter >= 5 {
+                logger.warn("STALE connection, force close")
+                self.close(isDisconnected: true) // call close function
+                zeroBytesOutCounter = 0 // reset counter
+            }
+        } else {
+            zeroBytesOutCounter = 0 // reset counter
+        }
```
shogo4405, line 472: During playback, there are cases where there is no data transfer for more than 5 seconds, which results in strong side effects. Due to this, it's not acceptable.

I would like this to be handled within the delegate method.
https://github.com/shogo4405/HaishinKit.swift/blob/main/Sources/RTMP/RTMPConnection.swift#L478

shogo4405: I'm considering whether these parameters around here can be used for handling this situation.
https://developer.apple.com/documentation/network/nwprotocoltcp/options#3225423

Sources/RTMP/RTMPConnection.swift
```diff
@@ -455,12 +457,26 @@ open class RTMPConnection: EventDispatcher {
...
+        
         for stream in streams {
             stream.on(timer: timer)
         }
         if measureInterval <= previousQueueBytesOut.count {
             var total = 0
-            for i in 0..<previousQueueBytesOut.count - 1 where previousQueueBytesOut[i] < previousQueueBytesOut[i + 1] {
+            for i in 0..<previousQueueBytesOut.count - 1 where previousQueueBytesOut[i] < previousQueueBytesOut[i + 1] && previousQueueBytesOut[i + 1] != 0 {
```
shogo4405, line 479: > there's a bug that sends publishSufficientBWOccured event

You are absolutely right! Thanks.

Sources/RTMP/RTMPConnection.swift
mkrn: OK, sorry didn't consider playback. 
As for using NWProtocolTCP options,  keep in mind, they're being overwritten here:  https://github.com/shogo4405/HaishinKit.swift/blob/3db54f064f3d66c18ffa3163dd771a5aec2cff63/Sources/RTMP/RTMPNWSocket.swift#L25 so it's impossible to set parameters from the outside.
I've tried setting them inside but it didn't affect the hanging back then. But can actually work now. Imho these timeouts should be very short, about 5s

Sources/RTMP/RTMPConnection.swift
shogo4405: This value should originally represent the unsent data size for a period of 1 second, so it seemed reasonable to consider it as sufficient when it's 0. However, due to it being synonymous with the size of the most recently enqueued data, issues were occurring. This has been fixed in this pull request: https://github.com/shogo4405/HaishinKit.swift/pull/1264.

Sources/RTMP/RTMPConnection.swift
shogo4405: I investigated, and it seems that the value of options.connectionDropTime is related to the switch from WiFi to Cellular. However, in any case, viabilityDidChange(to:) is always triggered, and I have confirmed that handling this method allows for proper closure without specifically checking connectionDropTime.

```diff
diff --git a/Sources/RTMP/RTMPConnection.swift b/Sources/RTMP/RTMPConnection.swift
index f89bc609..e9132df5 100644
--- a/Sources/RTMP/RTMPConnection.swift
+++ b/Sources/RTMP/RTMPConnection.swift
@@ -458,6 +458,7 @@ open class RTMPConnection: EventDispatcher {
         for stream in streams {
             stream.on(timer: timer)
         }
+        print(currentBytesOutPerSecond)
         if measureInterval <= previousQueueBytesOut.count {
             var total = 0
             for i in 0..<previousQueueBytesOut.count - 1 where previousQueueBytesOut[i] < previousQueueBytesOut[i + 1] {
diff --git a/Sources/RTMP/RTMPNWSocket.swift b/Sources/RTMP/RTMPNWSocket.swift
index 9155b329..14c31185 100644
--- a/Sources/RTMP/RTMPNWSocket.swift
+++ b/Sources/RTMP/RTMPNWSocket.swift
@@ -74,6 +74,27 @@ final class RTMPNWSocket: RTMPSocketCompatible {
         totalBytesOut.mutate { $0 = 0 }
         queueBytesOut.mutate { $0 = 0 }
         inputBuffer.removeAll(keepingCapacity: false)
+
+        let options = NWProtocolTCP.Options()
+        print(options.enableKeepalive)
+        options.connectionTimeout = 5
+        options.connectionDropTime = 10
+        options.persistTimeout = 15
+
+        print(parameters.debugDescription)
+
+        let tls = NWProtocolTLS.Options()
+        sec_protocol_options_set_verify_block(tls.securityProtocolOptions, { _, sec_trust, sec_protocol_verify_complete in
+            let trust = sec_trust_copy_ref(sec_trust).takeRetainedValue()
+            var error: CFError?
+            if SecTrustEvaluateWithError(trust, &error) {
+                sec_protocol_verify_complete(true)
+            } else {
+                sec_protocol_verify_complete(true)
+            }
+        }, self.networkQueue)
+
+        parameters = NWParameters(tls: tls, tcp: options)
         connection = NWConnection(to: NWEndpoint.hostPort(host: .init(withName), port: .init(integerLiteral: NWEndpoint.Port.IntegerLiteralType(port))), using: parameters)
         connection?.viabilityUpdateHandler = viabilityDidChange(to:)
         connection?.stateUpdateHandler = stateDidChange(to:)
```

shogo4405:
I am considering making changes to allow the definition of NWProtocolTCP.Options() from the outside in the future. Also, since the update process for statistics information has been fixed, I will close this PR. Thank you very much.

PR #1280, opened 2023-09-01
Improve audio channel mapping

leo150:
## Description & motivation

* Fix audio channel mapping. My previous implementation was creating a channel map of size n, where n is number of input channels. It should use number of output output channels instead.
* Expose channel map for external use
* Add channel map unit tests

## Type of change

### DIFF
--- a/Sources/Codec/AudioCodec.swift
+++ b/Sources/Codec/AudioCodec.swift
@@ -50,18 +50,13 @@ public class AudioCodec {
 
     /// Creates a channel map for specific input and output format
     static func makeChannelMap(inChannels: Int, outChannels: Int, outputChannelsMap: [Int: Int]) -> [NSNumber] {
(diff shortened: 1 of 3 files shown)


leo150:
I added unit tests. `makeChannelMap` falls back to default value if the provided one is incorrect.

shogo4405:
I just noticed after the merge that the tests are failing. Is the expected outcome correct?
スクリーンショット 2023-09-02 15 55 58

leo150:
Hey, thank you for noticing this. The test is incorrect, it should expect [0, -1]. I checked again, it's the only test that is failing. Thanks.

shogo4405:
Okay, I've fixed the tests. [2e1cb55](https://github.com/shogo4405/HaishinKit.swift/commit/2e1cb551795d3441ba842ca6223693dbe7ea5864)

PR #1281, opened 2023-09-01
Effects as array to execute them in known order

eerimoq:


eerimoq:
Why no merge? Anything more I have to do? Thanks.

shogo4405:
I couldn't evaluate the PR because it doesn't have any documentation content.

eerimoq:
Ok, I'll try to explain.

Registered video effects are currently stored in a set. Iterating over a set yields items in an undefined order. This PR replaces the set with an array to guarantee that the effects are executed in the same order as registered by the user.

eerimoq:
@shogo4405 Feel free to review again. Thanks!

eerimoq:
@shogo4405 Sorry for spam. I added documentation in the first PR comment. Would be happy if you can review again. Thanks.

levs42:
@shogo4405 could you please take a look at this PR? It's useful when the effects should be executed in a specific order. If Array isn't a good option, I can add [OrderedSet](https://github.com/apple/swift-collections/blob/main/Sources/OrderedCollections/OrderedSet/OrderedSet.swift) from Apple's collection. Thanks.

levs42:
@eerimoq could you please submit a new PR?

PR #1298, opened 2023-09-22
Support downmix feature and refactor.

shogo4405:
## Description & motivation
- I have added support for downmixing in multi-channel audio. Downmixing is a technique used to mix multiple channels into fewer channels, such as mixing down to 2 channels.
- I have corrected an issue where in PR #1280, when converting from 1 input channel to 2 output channels, it was mapping to the left channel. I have fixed this, and the expectation is to map to both channels now.

## Type of change


Sources/Media/IOAudioResampler.swift
```diff
@@ -10,13 +10,68 @@ protocol IOAudioResamplerDelegate: AnyObject {
...
+
+    init(sampleRate: Float64 = 0, channels: UInt32 = 0, downmix: Bool = false, channelMap: [NSNumber]? = nil) {
+        self.sampleRate = sampleRate
+        self.channels = channels
+        self.downmix = downmix
+        self.channelMap = channelMap
+    }
+
+    func invalidate(_ oldValue: IOAudioResamplerSettings!) -> Bool {
+        return (sampleRate != oldValue.sampleRate &&
```
levs42, line 27: Does this need to have OR condition? Please take look, thanks.

shogo4405: You're right. https://github.com/shogo4405/HaishinKit.swift/commit/780449bc2b36f2acaa3df1fdaf56ac65bc547f72

PR #1306, opened 2023-09-28
Restore AudioEffect feature

levs42:
## Description & motivation

Bringing AudioEffect back as it's very useful for exposing audio buffer from the library for an external use.

## Type of change


shogo4405:
I have removed it since it can be obtained through this delegate method. Since it shares the same interface as AudioEffect, I think it can be a suitable alternative.
https://github.com/shogo4405/HaishinKit.swift/blob/main/Sources/Net/NetStream.swift#L14

levs42:
Cool, I didn't see that method! Thank you for pointing to the right direction.

PR #1375, opened 2024-02-09
Change RTMPStream readyState observers to open

levs42:
## Description & motivation

This small change makes readyState open. It helps to observe changes of `readyState` property of `IOStream`.

## Type of change

### DIFF
--- a/Sources/IO/IOStream.swift
+++ b/Sources/IO/IOStream.swift
@@ -34,6 +34,10 @@ public protocol IOStreamDelegate: AnyObject {
     func stream(_ stream: IOStream, audioErrorOccurred error: IOAudioUnitError)
     /// Tells the receiver to the stream opened.
     func streamDidOpen(_ stream: IOStream)
+    /// Tells the receiver that the ready state will change.
+    func stream(_ stream: IOStream, willChangeReadyState state: IOStream.ReadyState)
+    /// Tells the receiver that the ready state did change.
+    func stream(_ stream: IOStream, didChangeReadyState state: IOStream.ReadyState)
 }
 
 @available(*, deprecated, renamed: "IOStream")
@@ -301,6 +305,7 @@ open class IOStream: NSObject {
             guard readyState != newValue else {
                 return
             }
+            delegate?.stream(self, willChangeReadyState: readyState)
             readyStateWillChange(to: newValue)
         }
         didSet {
@@ -308,6 +313,7 @@ open class IOStream: NSObject {
                 return
             }
             readyStateDidChange(to: readyState)
+            delegate?.stream(self, didChangeReadyState: readyState)
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/RTMP/RTMPStream.swift
```diff
@@ -416,7 +416,7 @@ open class RTMPStream: IOStream {
         return metadata
     }
 
-    override public func readyStateWillChange(to readyState: IOStream.ReadyState) {
+    override open func readyStateWillChange(to readyState: IOStream.ReadyState) {
```
shogo4405, line 419: If the goal is to detect changes in the readyState, I think it would be more iOS-like to provide a dedicated delegate method. Would achieving the requirements be difficult with such an approach?

levs42: Done, please take a look

Sources/IO/IOStream.swift
```diff
@@ -34,6 +34,10 @@ public protocol IOStreamDelegate: AnyObject {
     func stream(_ stream: IOStream, audioErrorOccurred error: IOAudioUnitError)
     /// Tells the receiver to the stream opened.
     func streamDidOpen(_ stream: IOStream)
+    /// Tells the receiver that the ready state will change.
+    func stream(_ stream: IOStream, willChange readyState: IOStream.ReadyState)
```
shogo4405, line 38: In Cocoa frameworks, abbreviations like "didChange" or "willChange" are not common, so could you please change them to "didReadyStateChange" or "willReadyStateChange"?

Sources/IO/IOStream.swift
```diff
@@ -458,6 +462,7 @@ open class IOStream: NSObject {
     /// A handler that receives stream readyState will update.
     /// - Warning: Please do not call this method yourself.
     open func readyStateWillChange(to readyState: ReadyState) {
+        delegate?.stream(self, willChange: readyState)
```
shogo4405, line 465: When it's located here, I tend to forget to implement it in the subclass. Therefore, I request the following changes.
```swift
public var readyState: ReadyState = .initialized {
    willSet {
        guard readyState != newValue else {
            return
        }
        delegate?.stream(self, willChange: newValue)
        readyStateWillChange(to: newValue)
    }
    didSet {
        guard readyState != oldValue else {
            return
        }
        readyStateDidChange(to: readyState)
        delegate?.stream(self, didChange: readyState)
    }
}
```

shogo4405 (changes requested):

Sources/IO/IOStream.swift
levs42: I've updated it to "willChangeReadyState" and "didChangeReadyState" as it sounds more readable. Please take a look.

shogo4405 (approved):
Thank you.

PR #1377, opened 2024-02-13
Add an option to show video output on a drawable without video effects

levs42:
## Description & motivation

This PR brings an option to display camera output on a drawable before applying the video effects. It allows to show the video effects only on the stream. Default value is the same as before. `effectsBuffer` flags helps to prevent the race condition when the video effects appear on the drawable for some frames; it happens when effects rendering is faster than drawable rendering.

## Type of change

### DIFF
--- a/Sources/IO/IOVideoMixerSettings.swift
+++ b/Sources/IO/IOVideoMixerSettings.swift
@@ -36,6 +36,8 @@ public struct IOVideoMixerSettings: Codable {
     public let direction: ImageTransform
     /// Specifies the main channel number.
     public var channel: UInt8 = 0
+    /// Specifies if effects are always rendered to a new buffer.
+    public var alwaysUseBufferPoolForVideoEffects: Bool = false
 
     /// Create a new IOVideoMixerSettings.
     public init(mode: Mode, cornerRadius: CGFloat, regionOfInterest: CGRect, direction: ImageTransform) {
--- a/Sources/IO/IOVideoMixer.swift
+++ b/Sources/IO/IOVideoMixer.swift
@@ -107,7 +107,7 @@ final class IOVideoMixer<T: IOVideoMixerDelegate> {
                 #if os(macOS)
                 pixelBufferPool?.createPixelBuffer(&imageBuffer)
                 #else
-                if buffer.width != Int(extent.width) || buffer.height != Int(extent.height) {
+                if settings.alwaysUseBufferPoolForVideoEffects || buffer.width != Int(extent.width) || buffer.height != Int(extent.height) {
                     pixelBufferPool?.createPixelBuffer(&imageBuffer)
                 }
                 #endif


shogo4405 (commented):
With HKView.isCaptureVideoPreviewEnabled = true, it's possible to preview the unprocessed video. Could this feature serve as an alternative?

levs42:
I tried `isCaptureVideoPreviewEnabled` but it doesn't lead to the same result. Here is what I'm trying to achieve with the PR:

<details>
  <summary>Screenshots</summary>
  
The preview shows two effects. First is time effect. Second is "view effect" of two yellow bars, a UIView is getting displayed on the screen and is getting rendered as a visual effect. The video preview displays both effects. The preview is in `resizeAspectFill` mode. This is the current behavior. 
IMG_9289

When `HKView.drawWithEffects = false` and `Stream.videoMixerSettings.effectsBuffer = true` both effects are not present on the preview:

IMG_9290
  
</details>

levs42:
I found the cause. Looks like `HKView.isCaptureVideoPreviewEnabled = true` works only for macos because it always [creates](https://github.com/shogo4405/HaishinKit.swift/blob/e85c8ec8c16d09dff329d46c74a98d744bb2816f/Sources/IO/IOVideoMixer.swift#L108) an extra buffer. I'm going to keep `effectsBuffer` flag for other platforms to be able to achieve the same behavior.

Sources/IO/IOVideoMixerSettings.swift
```diff
@@ -36,6 +36,8 @@ public struct IOVideoMixerSettings: Codable {
     public let direction: ImageTransform
     /// Specifies the main channel number.
     public var channel: UInt8 = 0
+    /// Specifies if effects are always rendered to a new buffer.
+    public var effectsBuffer: Bool = false
```
shogo4405, line 40: I see, understood. I accept changes of this magnitude. I'd like the variable name to be changed to "isXXXXEnabled". For example, "isEffectsBufferEnabled" is a bit unclear to me. Do you have any other suggestions for a more appropriate name?

levs42: We could use "alwaysUseBufferPoolForVideoEffects" or similar. "always" might be a good choice because even when the flag is false, macos will use the buffer pool. And when the flag is true, pixel buffer pool will always be used. Alternatively, we could add a new enum to the settings:
```
enum EffectsRenderMode {
   /// This mode tries to save memory on non-macos platforms by rendering to the video output pixel buffer.
   case optimized
   /// This mode is always using a separate pixel buffer to render the video effects.
   case separateBuffer
}
```

shogo4405: Let's go with "alwaysUseBufferPoolForVideoEffects".

levs42: Updated, please take a look

shogo4405 (approved):
👍  Thank you.

PR #1379, opened 2024-02-15
Validate audio channel map before applying

levs42:
## Description & motivation

Feature: Add audio channel map validation

Here's a case I'm trying to solve:
1. User connects an audio device with 2 input channels
2. User sets the output channel map to [0, 1]
3. User disconnects the input device, so now input has only 1 channel from the default mic
4. Audio converter crashes [here](https://github.com/shogo4405/HaishinKit.swift/blob/c712fe74d7fd555ada5b533dcdbfc939f4381caf/Sources/IO/IOAudioResampler.swift#L187) because it tries to map the second channel that isn't available anymore

There are other situations like this, for instance, just setting `stream.audioSettings.channelMap = [3]`.
The problem here is that audio input and output formats are not exposed outside of the library and `channelMap` cannot be corrected before it's applied to the converter. 

Theoretically, both formats or the converted itself could be exposed, but it has some downsides. For example, need to add extra delegate methods to IOAudioResamplerDelegate -> IOAudioUnitDelegate -> IOMixerDelegate -> IOStreamDelegate. So I choose an option to automatically validate the current channelMap to make sure that:
1. The length of the array is equal to the number of output channels
2. The array doesn't contain invalid input channels

## Type of change

### DIFF
--- a/Sources/IO/IOAudioResampler.swift
+++ b/Sources/IO/IOAudioResampler.swift
@@ -35,7 +35,7 @@ struct IOAudioResamplerSettings {
         if converter.downmix != downmix {
             converter.downmix = downmix
         }
-        if let channelMap {
+        if let channelMap = validatedChannelMap(converter) {
             converter.channelMap = channelMap
         } else {
             switch converter.outputFormat.channelCount {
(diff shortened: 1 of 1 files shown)


shogo4405:
Since crashes are not acceptable, I thought validation was necessary. For reference, since the input audio format is exposed via stream.audioInputFormat.

Issue #1384, opened 2024-03-08
App audio performance issue


levs42:
### Describe the bug

`IOAudioResampler.resample()` can't keep up with how fast ReplayKit sample provides `.audioApp` buffers. As result, `mixer.audioIO.lockQueue` keeps growing and broadcast extension exceeds 50MB RAM. Here's code example to reproduce the issue. I removed connect call, video and audioMic buffers for simplicity.

<details>
  <summary>SampleHandler</summary>
  
  ```swift


override open func broadcastStarted(withSetupInfo setupInfo: [String: NSObject]?) {
    LBLogger.with(HaishinKitIdentifier).level = .info
// rtmpConnection.connect(Preference.defaultInstance.uri!, arguments: nil)
    DispatchQueue.main.asyncAfter(deadline: .now() + 10) {
        self.isEnabled = true // delay to have time to connect cpu profiler before the crash
    }
}

override open func processSampleBuffer(_ sampleBuffer: CMSampleBuffer, with sampleBufferType: RPSampleBufferType) {
    guard isEnabled else { return }
    switch sampleBufferType {
    case .video:
        break
    case .audioMic:
        break
    case .audioApp:
        if CMSampleBufferDataIsReady(sampleBuffer) {
            rtmpStream.append(sampleBuffer)
        }
    @unknown default:
        break
    }
}
  ```
  
</details>

Here is resampler's audio formats on iPhone 11 Pro Max, iOS 17.4:
> inputFormat:<AVAudioFormat 0x3024bc8c0:  2 ch,  44100 Hz, Int16, interleaved>,outputFormat:<AVAudioFormat 0x3024bcbe0:  2 ch,  44100 Hz, Int16, interleaved>

<details>
  <summary>Profiler1</summary>

![Screenshot 2024-03-08 at 3 04 29 PM](https://github.com/shogo4405/HaishinKit.swift/assets/129108591/3f74ef70-0f16-436b-9033-d688854e02c6)
![Screenshot 2024-03-08 at 3 07 46 PM](https://github.com/shogo4405/HaishinKit.swift/assets/129108591/b61e8438-aa92-4d00-ad73-676eec6b168f)
![Screenshot 2024-03-08 at 11 29 44 AM](https://github.com/shogo4405/HaishinKit.swift/assets/129108591/bac239df-0e9f-47d9-8925-7e5aea7f40d0)
</details>

[...]

PR #1385, opened 2024-03-10
Implement timestamp recalculation for audio buffers

levs42:
## Description & motivation

This is a bugfix for #1384. Sometimes timescale of `CMSampleBuffer` isn't equal to the sample rate. For example, when ReplayKit sends `audioApp` sample buffers on iPhone 11 Pro Max with iOS 17.4. This PR adds an additional check to make sure timescale equals to the sample rate. If it does not, it will be recalculated like:
> new value = old value * sample rate / timescale

For example, a buffer with time stamp of:

> ▿ CMTime
>   - value : 44373327429041
>   - timescale : 1000000000

With current sampleRate 44100 will be recalculated to:
> ▿ CMTime
>   - value : 1956863739
>   - timescale : 44100

It might be good to keep the higher timescale for more accurate output timestamps, but it requires bigger rewrite of the ring buffer and resampler. 

## Type of change

### DIFF
--- a/Sources/IO/IOAudioResampler.swift
+++ b/Sources/IO/IOAudioResampler.swift
@@ -77,10 +77,6 @@ struct IOAudioResamplerSettings {
 final class IOAudioResampler<T: IOAudioResamplerDelegate> {
     private static func makeAudioFormat(_ inSourceFormat: inout AudioStreamBasicDescription) -> AVAudioFormat? {
         if inSourceFormat.mFormatID == kAudioFormatLinearPCM && kLinearPCMFormatFlagIsBigEndian == (inSourceFormat.mFormatFlags & kLinearPCMFormatFlagIsBigEndian) {
-            // ReplayKit audioApp.
-            guard inSourceFormat.mBitsPerChannel == 16 else {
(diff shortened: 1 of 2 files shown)


Sources/IO/IOAudioResampler.swift
```diff
@@ -77,10 +77,6 @@ struct IOAudioResamplerSettings {
 final class IOAudioResampler<T: IOAudioResamplerDelegate> {
     private static func makeAudioFormat(_ inSourceFormat: inout AudioStreamBasicDescription) -> AVAudioFormat? {
         if inSourceFormat.mFormatID == kAudioFormatLinearPCM && kLinearPCMFormatFlagIsBigEndian == (inSourceFormat.mFormatFlags & kLinearPCMFormatFlagIsBigEndian) {
-            // ReplayKit audioApp.
-            guard inSourceFormat.mBitsPerChannel == 16 else {
```
levs42, line 81: It isn't relevant to my audioApp audio format, but I removed it just in case.

shogo4405 (approved):
👍  Thanks.

PR #1388, opened 2024-03-13
Add a public method to send metadata

levs42:
## Description & motivation

This PR adds a public method to send the stream metadata manually. This allow to reset metadata timestamp and send metadata to the server when resolution/orientation of the stream has changed. It works on major platforms like YouTube and Twitch. Resetting the timestamp prevents video/audio desync. 

## Type of change

### DIFF
--- a/Sources/RTMP/RTMPStream.swift
+++ b/Sources/RTMP/RTMPStream.swift
@@ -371,6 +371,9 @@ open class RTMPStream: IOStream {
             guard let connection = self.connection, self.readyState == .publishing(muxer: self.muxer) else {
                 return
             }
+            if handlerName == "@setDataFrame" {
+                self.dataTimeStamps.removeValue(forKey: handlerName)
+            }
             let dataWasSent = self.dataTimeStamps[handlerName] == nil ? false : true
             let timestmap: UInt32 = dataWasSent ? UInt32((self.dataTimeStamps[handlerName]?.timeIntervalSinceNow ?? 0) * -1000) : UInt32(self.startedAt.timeIntervalSinceNow * -1000)
             let chunk = RTMPChunk(


shogo4405 (changes requested):
- I want to avoid convenient methods specific to services like Twitch because I can't maintain them.
- In this case, I think it would be sufficient to modify it so that when using rtmpStream.send @setMetaData, the timestamp can be sent as 0.

levs42:
Thank you for the review, this makes total sense. `send()` method is already available for the lib users for their convenience. I've removed new methods and added a check to clear timestamps for "@setDataFrame" message.

Also I'm not entirely sure because I don't use HK in playback mode, [this](https://github.com/shogo4405/HaishinKit.swift/blob/2e4171e4f96ec0bee51de70edd9fb235e63e1fd7/Sources/RTMP/RTMPMessage.swift#L431) place might be waiting for "@setDataFrame" as a handler name and "onMetaData" as an argument.

Sources/RTMP/RTMPStream.swift
```diff
@@ -371,6 +371,9 @@ open class RTMPStream: IOStream {
             guard let connection = self.connection, self.readyState == .publishing(muxer: self.muxer) else {
                 return
             }
+            if handlerName == "@setDataFrame" {
+                self.dataTimeStamps.removeValue(forKey: handlerName)
+            }
```
shogo4405, line 376: When considering Wowza, it becomes necessary to send timestamp differential. This will be marked as a change request as it could lead to a bug in Wowza.

shogo4405 (changes requested):

Sources/RTMP/RTMPStream.swift
levs42: When there is no value, the chunk type will be 0 which expects the absolute timestamp instead of delta. But looks like sending new absolute timestamp requires resetting some other values in the code. I will test Wowza next week. Maybe there could be a new argument in `send` specifying chunk type, it would also reset all timestamps.

shogo4405: Since the specification for @setDataFrame is undefined, it relies on server-side implementation. For the target Wowza, it was necessary to send it in this manner. However, it seems to not function correctly with nginx. If supporting Twitch behavior, providing an isResetTimestamp option seems like a good approach. https://github.com/shogo4405/HaishinKit.swift/pull/1390
https://github.com/shogo4405/HaishinKit.swift/blob/0a48aee6072f805537389e235b449934d06c7a34/Sources/RTMP/RTMPStream.swift#L368-L379

levs42: Yes this works well. I tested YouTube, Twitch and Wowza with `isResetTimestamp: true`. Can't verify nginx. Closing this PR.

levs42:
Closed in favor of #1390.

PR #1389, opened 2024-03-15
Add releaseStream command call

levs42:
## Description & motivation

This PR adds a `releaseStream` call that is used to release a published stream with the same name. This feature is undocumented but commonly used. For example, OBS [sends](https://github.com/obsproject/obs-studio/blob/7fab3d03e056405d4dbb5e2ed3a3f9b58e165876/plugins/obs-outputs/librtmp/rtmp.c#L3171) it before `FCPublish`. I thought to expose some interface so HK users could send it manually, however 1. HK already has undocumented `FCPublish`/`FCUnpublish` 2. When `releaseStream` isn't supported, the server will ignore it.

## Type of change

### DIFF
--- a/Sources/RTMP/RTMPStream.swift
+++ b/Sources/RTMP/RTMPStream.swift
@@ -256,6 +256,7 @@ open class RTMPStream: IOStream {
         addEventListener(.rtmpStatus, selector: #selector(on(status:)), observer: self)
         connection.addEventListener(.rtmpStatus, selector: #selector(on(status:)), observer: self)
         if connection.connected {
+            releaseStream()
             connection.createStream(self)
         }
         mixer.muxer = muxer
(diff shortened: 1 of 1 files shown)


levs42:
Looks like both OBS and [ffmpeg](https://ffmpeg.org/doxygen/0.6/rtmpproto_8c-source.html) (line 601) call `releaseStream` and `FCPublish` before `createStream`. Right now `FCPublish` is called after `createStream`. Should we move it?

shogo4405:
I cannot accept this because it references GPL source code. This decision is made to protect the businesses of HaishinKit users.

OBS is under the GPL license, and ffmpeg is under the LGPL license. Understanding that writing code while looking at these sources requires (L)GPL compliance.

Furthermore, it's understood that the AppStore is incompatible with GPL. In order to ensure everyone can conduct business on the AppStore with peace of mind, I maintain the BSD license for HaishinKit development. I develop HaishinKit without looking at the code of OBS or ffmpeg.

PR #1391, opened 2024-03-18
Add releaseStream support and fix FCPublish call order

levs42:
## Description & motivation

This PR allows the library users to be able to call code before `createStream`. For example, to call `releaseStream` which is supported by Wowza. New state is required because `.initialized` is the default state and when the connection is successful, changing the ready state [here](https://github.com/shogo4405/HaishinKit.swift/blob/2e4171e4f96ec0bee51de70edd9fb235e63e1fd7/Sources/RTMP/RTMPStream.swift#L550) doesn't lead to the delegate notification call. In addition, `createStream` is wrapped into a lock queue, so library users have time to perform extra calls on the same queue before the create stream execution. It also adds an additional `send` method to allow `connection.call` method call from the outside.

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/Sources/RTMP/RTMPConnection.swift
+++ b/Sources/RTMP/RTMPConnection.swift
@@ -179,6 +179,8 @@ public class RTMPConnection: EventDispatcher {
     public var chunkSize: Int = RTMPConnection.defaultChunkSizeS
     /// Specifies the URI passed to the Self.connect() method.
     public private(set) var uri: URL?
+    /// Specifies a name to pass to `releaseStream` command.
+    public private(set) var resourceName: String?
     /// Specifies the instance connected to server(true) or not(false).
     public private(set) var connected = false
     /// Specifies the instance requires Network.framework if possible.
@@ -282,11 +284,12 @@ public class RTMPConnection: EventDispatcher {
     }
 
     /// Creates a two-way connection to an application on RTMP Server.
-    public func connect(_ command: String, arguments: Any?...) {
+    public func connect(_ command: String, name: String? = nil, arguments: Any?...) {
         guard let uri = URL(string: command), let scheme = uri.scheme, !connected && Self.supportedProtocols.contains(scheme) else {
             return
         }
         self.uri = uri
+        self.resourceName = name
         self.arguments = arguments
         switch scheme {
         case "rtmpt", "rtmpts":
@@ -338,6 +341,9 @@ public class RTMPConnection: EventDispatcher {
     }
 
     func createStream(_ stream: RTMPStream) {
+        if let resourceName {
+            call("releaseStream", responder: nil, arguments: resourceName)
+        }
         let responder = RTMPResponder(result: { data -> Void in
             guard let id = data[0] as? Double else {
                 return
@@ -386,7 +392,7 @@ public class RTMPConnection: EventDispatcher {
                 break
             case description.contains("reason=needauth"):
                 let command = Self.makeSanJoseAuthCommand(uri, description: description)
-                connect(command, arguments: arguments)
+                connect(command, name: resourceName, arguments: arguments)
             case description.contains("authmod=adobe"):
                 if user.isEmpty || password.isEmpty {
                     close(isDisconnected: true)
@@ -394,7 +400,7 @@ public class RTMPConnection: EventDispatcher {
                 }
                 let query = uri.query ?? ""
                 let command = uri.absoluteString + (query.isEmpty ? "?" : "&") + "authmod=adobe&user=\(user)"
-                connect(command, arguments: arguments)
+                connect(command, name: resourceName, arguments: arguments)
             default:
                 break
             }
--- a/Sources/RTMP/RTMPConnection.swift
+++ b/Sources/RTMP/RTMPConnection.swift
@@ -343,6 +343,7 @@ public class RTMPConnection: EventDispatcher {
     func createStream(_ stream: RTMPStream) {
         if let resourceName {
             call("releaseStream", responder: nil, arguments: resourceName)
+            call("FCPublish", responder: nil, arguments: resourceName)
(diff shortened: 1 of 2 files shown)


Sources/RTMP/RTMPStream.swift
```diff
@@ -389,6 +389,11 @@ open class RTMPStream: IOStream {
         }
     }
 
+    /// Sends a command to the RTMP client or server.
+    public func send(commandName: String, arguments: Any?...) {
```
shogo4405, line 393: [IMO]
RTMPConneciton.call is already public, so I don't feel the need for implementation on the HaishinKit side. We expect it to be defined in the importing side's extension or RTMPStream extension.

Sources/IO/IOStream.swift
```diff
@@ -57,6 +57,8 @@ open class IOStream: NSObject {
         }
 
         /// NetStream has been created.
+        case uninitialized
```
shogo4405, line 60: With the increase of "uninitialized," the number of states in readyState has increased, making it difficult to foresee and maintain. I also think "closed = uninitialized" is synonymous.

shogo4405 (commented):
Upon reconsideration of FCPublish/FCUnpublish, it seems more natural for them to exist in RTMPConnection rather than RTMPStream. I believe it would be more appropriate to implement them properly on the RTMPConnection side rather than modifying RTMPStream.

Even if it were to be merged, I have no plans to accept it, as the relevant code itself will be removed in the next version.

Sources/RTMP/RTMPStream.swift
levs42: Sounds good. `connection` is private, but it's injected into the `RTMPStream` so it can be used from the outside.

Sources/IO/IOStream.swift
shogo4405: If you modify createStream as follows, it ensures that the call to createStream will occur after the releaseStream command is received.
```
 func createStream(_ stream: RTMPStream) {
        let releaseStreamResponder = RTMPResponder(result: { data -> Void in
                let responder = RTMPResponder(result: { data -> Void in
            guard let id = data[0] as? Double else {
                return
            }
            stream.id = UInt32(id)
            stream.readyState = .open
           })
           call("createStream", responder: responder)
        })
        call(“releaseStream”, responder: releaseStreamResponder, arguments: stream.key)
    }
```
In modern times, it cannot be simply assumed whether various server-side implementations are correctly implemented or not. I think it would be good to call it like this.
```
func createStream(_ stream: RTMPStream) {
    // Specify an empty responder to ignore the result of releaseStream.
    let releaseStreamResponder = RTMPResponder(result: { data -> Void in })
    call(“releaseStream”, responder: releaseStreamResponder, arguments: stream.key)
    call(“createStream”, responder: …)
}
```

levs42: I've updated the PR with optional releaseStream call without a responder.

levs42:
In addition to optional `releaseStream` call, I moved `FCPublish` to be called before `createStream`. I can reference Adobe software and other encoders that encode in this order. In addition, Twitter doesn't respond with `onFCPublish` when `FCPublish` is after `createStream`:

<details>
  <summary>Screenshot</summary>
  
Screenshot 2024-03-20 at 3 03 12 PM
  
</details>

Now it's a breaking change because users have to pass `name` to `connect()` method in order to `FCPublish` to work.

Sources/RTMP/RTMPConnection.swift
```diff
@@ -282,11 +284,12 @@ public class RTMPConnection: EventDispatcher {
     }
 
     /// Creates a two-way connection to an application on RTMP Server.
-    public func connect(_ command: String, arguments: Any?...) {
+    public func connect(_ command: String, name: String? = nil, arguments: Any?...) {
```
shogo4405, line 287: [MUST]
I don't want to make such disruptive changes for the purpose of supporting releaseStream. Also, considering viewing or multiple streams, I still think this change is unnecessary.

If I was to do it, I think it would be better to prepare an RTMPConnection subclass or wrapper class, such as RTMPConnectionFMLECompatible, that conforms to the FMLE sequence.

shogo4405 (changes requested):

Sources/RTMP/RTMPConnection.swift
levs42: Should I make RTMPConnection and `createStream()` open to allow subclassing? My goal is to have any kind of solution for this case being supported. It's common and solves platform specific issues while not causing others because all other popular encoders perform it in the mentioned sequence. FMLE app performs it in the following order:

> AMF command 'releaseStream'
> AMF command 'FCPublish'
> AMF command 'createStream'

shogo4405: Yes. HaishinKit aims to provide multi-streaming and ingest/playback functionalities. Killing off RTMPConnection functionality or making destructive changes to methods in pursuit of FMLE compatibility would diverge from this objective.

Therefore, I believe it would be appropriate to have a class dedicated to single-stream ingest, inheriting from RTMPConnection, and within that class, call the FMLE-compatible methods.

shogo4405: It seems natural for the stream name of releaseStream to be in RTMPStream.
To avoid issues with multi-streaming and playback without inheriting RTMPConnection, this implementation seems necessary.
https://github.com/shogo4405/HaishinKit.swift/pull/1393/files

It might be good to handle errors when there is a contradiction between the publish and fcPublishName, but I don't think it's necessary to be that strict.

shogo4405:
Close as https://github.com/shogo4405/HaishinKit.swift/pull/1393

Issue #758, opened 2020-09-27
Connection Refused (iPhone -> Nginx on Localhost)


Taioron:
Getting connection refused errors. Anyone see anything? (Server config at the bottom)
```
2020-27-09 11:32:37.216 [Info] [com.haishinkit.HaishinKit] [VideoIOComponent.swift:96] fps > (fps: 30.0, duration: __C.CMTime(value: 100, timescale: 3000, flags: __C.CMTimeFlags(rawValue: 1), epoch: 0))
2020-09-27 11:32:37.233522-0500 MyApp[931:151398] [connection] nw_socket_handle_socket_event [C6:1] Socket SO_ERROR [61: Connection refused]
2020-09-27 11:32:37.234506-0500 MyApp[931:151267] [connection] nw_connection_get_connected_socket [C6] Client called nw_connection_get_connected_socket on unconnected nw_connection
2020-09-27 11:32:37.234615-0500 MyApp[931:151267] TCP Conn 0x2811a4a50 Failed : error 0:61 [61]
```

**Xcode:**

```
   @IBAction func streamAction(_ sender: Any) {
        startAudioSession()
        setupRTMPStream()
    }
    
    
    func startAudioSession(){
    let session = AVAudioSession.sharedInstance()
    do {
        // https://stackoverflow.com/questions/51010390/avaudiosession-setcategory-swift-4-2-ios-12-play-sound-on-silent
        if #available(iOS 10.0, *) {
            try session.setCategory(.playAndRecord, mode: .default, options: [.defaultToSpeaker, .allowBluetooth])
        } else {
            session.perform(NSSelectorFromString("setCategory:withOptions:error:"), with: AVAudioSession.Category.playAndRecord, with: [
                AVAudioSession.CategoryOptions.allowBluetooth,
                AVAudioSession.CategoryOptions.defaultToSpeaker]
            )
            try session.setMode(.default)
        }
        try session.setActive(true)
    } catch {
        print(error)
    }
    }
    
    func setupRTMPStream(){
        mapView.isHidden = true
        streamOutlet.isHidden = false
        let rtmpConnection = RTMPConnection()
        let rtmpStream = RTMPStream(connection: rtmpConnection)
        rtmpStream.captureSettings = [
            .fps: 30, // FPS
            .sessionPreset: AVCaptureSession.Preset.medium, // input video width/height
            // .isVideoMirrored: false,
            // .continuousAutofocus: false, // use camera autofocus mode
[...]

PR #1447, opened 2024-05-20
README: add instructions for running the examples

PBHDK:
## Description & motivation

To make it easier for newcomers, like myself, to get up and running with the HaishinKit example apps, expand on the setup instructions.

The nginx configuration recommendation comes from: https://github.com/shogo4405/HaishinKit.swift/issues/758

## Type of change


shogo4405:
Thank you for the PR.

I take the stance that users should prepare their own servers, so I do not plan to include detailed instructions in the README. Including it in the README would lead to questions about brew installation and build errors, which I cannot support.

Since I will not accept the PR itself, close it.

Issue #481, opened 2018-11-23
Stream is always portrait in Broadcast Upload Extension


arslan2012:
Stream is always portrait in Broadcast Upload Extension

I used the exact copy of Examples/iOS/Screencast

When I enter games that are landscape I expect the output to be landscape as well. But it's not.

I tried
```swift
broadcaster.stream.syncOrientation = true
// And
broadcaster.stream.orientation = .landscapeLeft
```

both does not work

shogo4405:
Yes. Always portrait :(
It seems this behavior is a specification.

arslan2012:
Isn't there anyway that can rotate the stream?

arslan2012:
ok I made it work by adding a custom effect

```swift
final class RotationEffect: VisualEffect {
    override func execute(_ image: CIImage, info: CMSampleBuffer?) -> CIImage {
        guard #available(iOS 11.0, *),
              let info = info,
              let orientationAttachment = CMGetAttachment(info, key: RPVideoSampleOrientationKey as CFString, attachmentModeOut: nil) as? NSNumber,
              let orientation = CGImagePropertyOrientation(rawValue: orientationAttachment.uint32Value) else {
            return image
        }
        switch orientation {
        case .left:
            return image.oriented(.right)
        case .right:
            return image.oriented(.left)
        default:
            return image
        }
    }
}
```

but because RPVideoSampleOrientationKey is only available in ios11 or newer. this does not work on ios10, is there any other way we can get orientation in an VisualEffect? @shogo4405

locorocoo:
@shogo4405 Any update on this? I'm also experiencing similar issues

shogo4405:
@locorocoo No. I won't fix this issue.

Ankush-Redapple:
@shogo4405 Any update on this? We many developers love this library. It's truly great and save lots of time. Could please help out regarding orientation?

mooncatventures-group:
> @locorocoo No. I won't fix this issue.

The fact that this library doesn't handle orientation changes to landscape, and freezes when you rotate the device makes it unusable for replaykit 2.

joshuadiezmo:
> ok I made it work by adding a custom effect
> 
> ```swift
> final class RotationEffect: VisualEffect {
>     override func execute(_ image: CIImage, info: CMSampleBuffer?) -> CIImage {
>         guard #available(iOS 11.0, *),
>               let info = info,
[...]

PR #1452, opened 2024-05-24
Implement orientation change handling for ReplayKit

levs42:
## Description & motivation

This PR adds a feature to rotate pixel buffers that come from ReplayKit. This allows to rotate video when device changes orientation. It is much more efficient than using `VideoEffect`. 

Rotation has two modes besides default `.disabled`.
####  Automatic:
```Swift
rtmpStream.videoMixerSettings.propogateReplayKitOrientation = true
rtmpStream.videoSettings.rotationMode = .automatic
```
In this case `VideoCodec` automatically determines the required rotation based on the buffer orientation.

#### Manual:
```Swift
rtmpStream.videoSettings.rotationMode = ._90
```

This allows HK users to specify rotation manually.

## Notes
* I decided to store rotation information in the `CVImageBuffer` so there is no need to modify `VideoCodec` method signatures. As result, `IOVideoMixer` saves the orientation info as an attachment of the pixel buffer. It is only required for `.automatic` mode.
* `CVPixelBufferPool` leaks in iOS 17.5. When pool reference is set to `nil`, it keeps circular reference to itself and allocated pixel buffers are also getting retained. That's why I'm not using pool to allocate `rotationPixelBuffer`. 

## Type of change

### DIFF
--- a/Sources/Screen/VideoRotator.swift
+++ b/Sources/Screen/VideoRotator.swift
@@ -1,0 +1,173 @@
+import CoreImage
+import Foundation
+import ReplayKit
+import VideoToolbox
+
+public enum VideoRotatorError: Error {
+    case noImageBuffer
+    case noOrientationInfo
+    case unsupportedOrientation
+    case cannotAllocatePixelBuffer(CVReturn)
+    case rotationFailure(OSStatus)
+}
+
+public class VideoRotator {
+    private var pixelInfo = PixelInfo.zero {
+        didSet {
+            guard pixelInfo != oldValue else {
+                return
+            }
+            rotationPixelBuffer = nil
+            pixelBufferStatus = CVPixelBufferCreate(
+                kCFAllocatorDefault,
+                pixelInfo.width,
+                pixelInfo.height,
+                pixelInfo.format,
+                nil,
+                &rotationPixelBuffer
+            )
+        }
+    }
+    private var rotationPixelBuffer: CVPixelBuffer?
+    private var pixelBufferStatus: CVReturn = kCVReturnSuccess
+    private let session: VTPixelRotationSession
+
+    public init?() {
+        guard #available(iOS 16.0, tvOS 16.0, macOS 13.0, *) else {
+            return nil
+        }
+        var session: VTPixelRotationSession?
+        let status = VTPixelRotationSessionCreate(kCFAllocatorDefault, &session)
+        guard status == noErr, let session else {
+            return nil
+        }
+        self.session = session
+    }
+
+    @available(iOS 16.0, tvOS 16.0, macOS 13.0, *)
+    public func rotate(buffer sampleBuffer: CMSampleBuffer) -> Result<CMSampleBuffer, VideoRotatorError> {
+        guard let buffer = sampleBuffer.imageBuffer else {
+            return .failure(.noImageBuffer)
+        }
+        buffer.lockBaseAddress()
+        defer {
+            buffer.unlockBaseAddress()
+        }
+        var orientation: CGImagePropertyOrientation?
+        orientation = sampleBuffer.orientation
+        guard let orientation else {
+            return .failure(.noOrientationInfo)
+        }
+        guard orientation != .up else {
+            return .success(sampleBuffer)
+        }
+        var status: OSStatus
+        switch orientation {
+        case .left:
+            status = session.setOption(.init(key: .rotation, value: ._90))
+        case .down:
+            status = session.setOption(.init(key: .rotation, value: ._180))
+        case .right:
+            status = session.setOption(.init(key: .rotation, value: ._270))
+        default:
+            return .failure(.unsupportedOrientation)
+        }
+        guard status == noErr else {
+            return .failure(.rotationFailure(status))
+        }
+        switch orientation {
+        case .up, .down:
+            pixelInfo = .init(width: buffer.width, height: buffer.height, format: buffer.formatType)
+        case .left, .right:
+            pixelInfo = .init(width: buffer.height, height: buffer.width, format: buffer.formatType)
+        default:
+            return .failure(.unsupportedOrientation)
+        }
+        guard let rotationPixelBuffer else {
+            return .failure(.cannotAllocatePixelBuffer(pixelBufferStatus))
+        }
+        status = VTPixelRotationSessionRotateImage(session, buffer, rotationPixelBuffer)
+        guard status == noErr else {
+            return .failure(.rotationFailure(status))
+        }
+        var rotatedBuffer: CMSampleBuffer?
+        (rotatedBuffer, status) = createSampleBuffer(sampleBuffer, imageBuffer: rotationPixelBuffer)
+        guard let rotatedBuffer else {
+            return .failure(.rotationFailure(status))
+        }
+        return .success(rotatedBuffer)
+    }
+
+    @inline(__always)
+    private func createSampleBuffer(_ inSampleBuffer: CMSampleBuffer, imageBuffer: CVImageBuffer) -> (CMSampleBuffer?, OSStatus) {
+        var info = CMSampleTimingInfo()
+        info.presentationTimeStamp = inSampleBuffer.presentationTimeStamp
+        info.duration = CMSampleBufferGetOutputDuration(inSampleBuffer)
+        info.decodeTimeStamp = CMSampleBufferGetDecodeTimeStamp(inSampleBuffer)
+        var formatDescription: CMFormatDescription?
+        var status = CMVideoFormatDescriptionCreateForImageBuffer(
+            allocator: kCFAllocatorDefault,
+            imageBuffer: imageBuffer,
+            formatDescriptionOut: &formatDescription)
+        guard status == noErr, let formatDescription else {
+            return (nil, status)
+        }
+        var outSampleBuffer: CMSampleBuffer?
+        status = CMSampleBufferCreateReadyWithImageBuffer(allocator: kCFAllocatorDefault,
+                                                          imageBuffer: imageBuffer,
+                                                          formatDescription: formatDescription,
+                                                          sampleTiming: &info,
+                                                          sampleBufferOut: &outSampleBuffer)
+
+        guard status == noErr, let outSampleBuffer else {
+            return (nil, status)
+        }
+        return (outSampleBuffer, noErr)
+    }
+}
+
+private struct PixelInfo: Hashable, Equatable, CustomStringConvertible {
+    static let zero: PixelInfo = .init(width: 0, height: 0, format: 0)
+
+    let width: Int
+    let height: Int
+    let format: OSType
+
+    static func == (lhs: PixelInfo, rhs: PixelInfo) -> Bool {
+        return lhs.width == rhs.width && lhs.height == rhs.height && lhs.format == rhs.format
+    }
+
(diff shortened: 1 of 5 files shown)


shogo4405:
I tested it on my end, but it seems that playback with ffplay does not work properly.
スクリーンショット 2024-05-26 6 39 54

shogo4405:
I wasn't able to test it, so I need to confirm—could you explain specifically what kind of feature this is? How will it affect the user experience?

Since the video mixing functionality will undergo a design change in the next version, I think it would be better to implement it with the new design. https://github.com/shogo4405/HaishinKit.swift/pull/1449

levs42:
Could you please share ffplay command that I could use to reproduce the green screen issue?

As for the feature, the main purpose is to be able to rotate pixel buffers when user is screen sharing on iOS. Old behavior: user starts a stream in a portrait mode; rotates the phone horizontally; the video stays vertical. New behaviour: user rotates the phone to landscape; video becomes _horizontal_. 

Here's an issue: #481. Some screenshots:

<details>
  <summary>User starts a stream in portrait mode</summary>
  
Screenshot 2024-05-25 at 6 33 07 PM

</details>

<details>
  <summary>Old: user rotates phone horizontally</summary>
  
Screenshot 2024-05-25 at 6 31 27 PM

</details>

<details>
  <summary>New: user rotates phone horizontally</summary>
  
Screenshot 2024-05-25 at 6 31 46 PM


</details>

I was hoping to be able to merge this before the redesign if I fix the issues, I don't mind updating it too.

shogo4405:
`ffplay rtmp://localhost/live/live`

```sh
home HaishinKit.swift % ffmpeg -version
ffmpeg version 6.1.1 Copyright (c) 2000-2023 the FFmpeg developers
built with Apple clang version 15.0.0 (clang-1500.3.9.4)
configuration: --prefix=/opt/homebrew/Cellar/ffmpeg/6.1.1_6 --enable-shared --enable-pthreads --enable-version3 --cc=clang --host-cflags= --host-ldflags='-Wl,-ld_classic' --enable-ffplay --enable-gnutls --enable-gpl --enable-libaom --enable-libaribb24 --enable-libbluray --enable-libdav1d --enable-libharfbuzz --enable-libjxl --enable-libmp3lame --enable-libopus --enable-librav1e --enable-librist --enable-librubberband --enable-libsnappy --enable-libsrt --enable-libssh --enable-libsvtav1 --enable-libtesseract --enable-libtheora --enable-libvidstab --enable-libvmaf --enable-libvorbis --enable-libvpx --enable-libwebp --enable-libx264 --enable-libx265 --enable-libxml2 --enable-libxvid --enable-lzma --enable-libfontconfig --enable-libfreetype --enable-frei0r --enable-libass --enable-libopencore-amrnb --enable-libopencore-amrwb --enable-libopenjpeg --enable-libopenvino --enable-libspeex --enable-libsoxr --enable-libzmq --enable-libzimg --disable-libjack --disable-indev=jack --enable-videotoolbox --enable-audiotoolbox --enable-neon
```

shogo4405:
https://github.com/shogo4405/HaishinKit.swift/blob/e4b26176942956760df826e766902880879dd60b/Tests/ISO/AVCDecoderConfigurationRecordTests.swift#L7-L20

When it works as intended, the avcc looks like the example above. When it doesn't work, it turns into the values in the example below. It seems the sequenceParameterSets value is insufficient. I verified this with nginx + rtmp-module.

**testMain_1() print**
```
AVCDecoderConfigurationRecord{configurationVersion: 1,avcProfileIndication: 66,profileCompatibility: 0,avcLevelIndication: 40,lengthSizeMinusOneWithReserved: 255,numOfSequenceParameterSetsWithReserved: 225,sequenceParameterSets: [[39, 66, 0, 40, 171, 64, 60, 3, 143, 83, 77, 192, 128, 128, 128, 128]],pictureParameterSets: [[40, 206, 60, 128]],chromaFormatWithReserve: 0,bitDepthLumaMinus8WithReserve: 0,bitDepthChromaMinus8WithReserve: 0,sequenceParameterSetExt: []}
```
**testMain_2() print**
```
AVCDecoderConfigurationRecord{configurationVersion: 1,avcProfileIndication: 66,profileCompatibility: 0,avcLevelIndication: 40,lengthSizeMinusOneWithReserved: 255,numOfSequenceParameterSetsWithReserved: 225,sequenceParameterSets: [[39, 66, 0, 40, 171, 64, 60, 3, 143, 83, 32]],pictureParameterSets: [[40, 206, 60, 128]],chromaFormatWithReserve: 0,bitDepthLumaMinus8WithReserve: 0,bitDepthChromaMinus8WithReserve: 0,sequenceParameterSetExt: []}
```

levs42:
I've updated the PR and the green screen issue should be fixed. Looks like danger cannot see the commit from our fork. Please take a look.

shogo4405:
I tested it on my end, but it seems to be fixed in landscape mode. Are there any special settings?

I understand that many services assume the resolution does not change during streaming. I wanted to verify this because changing the resolution could result in unpredictable behavior, potentially affecting recordings and HLS viewing.

Since these features are specialized for ReplayKit, I thought it might be difficult to maintain if modifications were made to IOVideoMixer or VideoCodec. The following design should be acceptable. Please consider it.

## Proposal 1

- Create a class specialized in rotating the CMSampleBuffer class from ReplayKit.
- Use this class to rotate the sampleBuffer and pass it to `stream.append(sampleBuffer)`.
It will be necessary to detect the format before passing it to the videoCodec, so prepare a delegate specialized for this.

## Proposal 2*

- From the next version, it will be possible to specify any resolution for streaming, regardless of the input CMSampleBuffer.
- Using this mechanism, make it possible to rotate and set the screen rotation metadata from ReplayKit.

P.S.
I will check the issue with Danger when I have time.

levs42:
> I tested it on my end, but it seems to be fixed in landscape mode. Are there any special settings?

I added `isPortrait = false` in SampleHandler so it would be easier to control video size. However it might be confusing so I removed it. Now the video will stay at portrait size even for landscape video unless specified with `videoSize` as before.

> changing the resolution could result in unpredictable behavior
Rotation doesn't change the output resolution. The rotated buffer is cropped into target `videoSize` and scaled based on `scalingMode`.

Thank you for the feedback regarding the possible design changes. Users can provide rotation for non-ReplayKit use cases too. Also based on the provided details could you please reconsider the proposals? Thanks.

shogo4405:
I think method 1 is good because it can be done as an add-on without modifying HaishinKit itself. It looks something like this.
```
protocol CMSampleBufferRotaterDelegate {
    func rotater(_ rotater: CMSampleBufferRotater, didRotate sampleBuffer: CMSampleBuffer)
}

class CMSampleBufferRotater {
    weak var delegate: 

    // any options
 
    func append(_ sampleBuffer: CMSampleBuffer) {
    // Rotation processing
    delegate?.rotaer(self, sampleBuffer)
    }
}

SampleHandler {
    var rotater: CMSampleBufferRotater

     override open func processSampleBuffer(_ sampleBuffer: CMSampleBuffer, with sampleBufferType: RPSampleBufferType) {
        switch sampleBufferType {
        case .video:
            rotater.append(sampleBuffer)
        case .audioMic:
            Break
     }
}

extension SampleHandler: CMSampleBufferRotaterDelegate {
     func rotater(_ rotater: CMSampleBufferRotater, didRotate sampleBuffer: CMSampleBuffer) {
          rtmpStream.append(sampleBuffer)
     }
}
```

levs42:
Library users won't be able to do screen sharing on iOS without correctly rotating the video buffers that's why I thought to include it directly into HaishinKit.

levs42:
I've implemented a dedicated video rotation class and updated the ReplayKit example. Please take a look.

While working on it I realized that `RPVideoSampleOrientationKey` is part of the attachments of the buffer that is supplied to the library. Therefore it should be expected for the attachments to be handled inside of the library. I hope this code can be part of `IOVideoMixer` in the future versions.

Sources/Screen/VideoRotator.swift
```diff
@@ -0,0 +1,173 @@
+import CoreImage
+import Foundation
+import ReplayKit
+import VideoToolbox
+
```
shogo4405, line 5: [MUST]
* Please include documentation for the public API.

Sources/Screen/VideoRotator.swift
```diff
@@ -0,0 +1,173 @@
...
+public enum VideoRotatorError: Error {
+    case noImageBuffer
+    case noOrientationInfo
+    case unsupportedOrientation
+    case cannotAllocatePixelBuffer(CVReturn)
+    case rotationFailure(OSStatus)
+}
+
+public class VideoRotator {
```
shogo4405, line 14: [MUST]
* Please include documentation for the public API.

Sources/Screen/VideoRotator.swift
```diff
@@ -0,0 +1,173 @@
...
+        }
+    }
+    private var rotationPixelBuffer: CVPixelBuffer?
+    private var pixelBufferStatus: CVReturn = kCVReturnSuccess
+    private let session: VTPixelRotationSession
+
+    public init?() {
+        guard #available(iOS 16.0, tvOS 16.0, macOS 13.0, *) else {
+            return nil
+        }
```
shogo4405, line 38: [MUST]
- Please mark the class itself with @available(iOS 16.0, tvOS 16.0, macOS 13.0, *) instead of adding it to each method.

Sources/Screen/VideoRotator.swift
```diff
@@ -0,0 +1,173 @@
...
+
+    static func == (lhs: PixelInfo, rhs: PixelInfo) -> Bool {
+        return lhs.width == rhs.width && lhs.height == rhs.height && lhs.format == rhs.format
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(width)
+        hasher.combine(height)
+        hasher.combine(format)
+    }
```
shogo4405, line 144: [IMO]
- Let's remove it. If it's a private struct, adding Hashable and Equatable will automatically generate the relevant code. Let Xcode handle it instead of writing it yourself.

shogo4405 (changes requested):
Thank you. It was as expected.
There are a few parts that need to be addressed, so I have marked it as Request Changes.

levs42:
Thanks for the comments! I updated the code

shogo4405 (approved):
👍

shogo4405:
[FYI] The sample code didn't work. It worked when I did it like this.
```diff
diff --git a/Examples/iOS/Screencast/SampleHandler.swift b/Examples/iOS/Screencast/SampleHandler.swift
index 26013f5e..2633c15d 100644
--- a/Examples/iOS/Screencast/SampleHandler.swift
+++ b/Examples/iOS/Screencast/SampleHandler.swift
@@ -12,8 +12,15 @@ open class SampleHandler: RPBroadcastSampleHandler {
     private var _rotator: Any?
     @available(iOS 16.0, tvOS 16.0, macOS 13.0, *)
     private var rotator: VideoRotator? {
-        get { _rotator as? VideoRotator }
-        set { _rotator = newValue }
+        get {
+            if _rotator == nil {
+                _rotator = VideoRotator()
+            }
+            return _rotator as? VideoRotator
+        }
+        set {
+            _rotator = newValue
+        }
     }
```

levs42:
Nice catch! Thank you

PR #1465, opened 2024-06-07
Change how track volume is applied

levs42:
## Description & motivation

When `settings` are applied to `IOAudioMixerByMultiTrack` before it receives an audio buffer, `mixerNode` is not initialized. This fix applies the volume during the node creation. Second issue is related to using non-sequential track numbers. For example, 0 for microphone, 2 for screensharing app audio. This results to only tracks 0, 1 being enabled and 2 being disabled. 

## Type of change


shogo4405 (approved):
👍

PR #1499, opened 2024-06-26
Fix compiling issue in Xcode16 Beta

RainyTunes:
## Description & motivation
There is a small compiling issue after I update to the newest verstion Xcode 16 Beta, and this is the autofix suggestion of Xcode compiler.

## Type of change


shogo4405:
The issue has been fixed in version 1.9.0. I believe it would be beneficial to apply a patch to the 1.8.2 series, so I will proceed with that. Handling this as a PR would be complicated, so please close this as is. Thank you.

shogo4405:
I have fixed it as version 1.8.3. Thank you for the feedback. https://github.com/shogo4405/HaishinKit.swift/releases/tag/1.8.3

PR #1500, opened 2024-06-28
Update audio track delegate on output format change

levs42:
## Description & motivation

This PR fixes an issue where audio tracks from multitrack mixer don't preserve their delegates. This leads to a state where AudioCodec doesn't receive current audio format from the mixer.

## Type of change

### DIFF
--- a/Sources/IO/IOAudioMixerByMultiTrack.swift
+++ b/Sources/IO/IOAudioMixerByMultiTrack.swift
@@ -27,6 +27,7 @@ final class IOAudioMixerByMultiTrack: IOAudioMixerConvertible {
             for id in tracks.keys {
                 buffers[id] = .init(outputFormat)
                 tracks[id] = .init(id: id, outputFormat: outputFormat)
+                tracks[id]?.delegate = self
             }
         }
     }


Sources/IO/IOAudioMixerByMultiTrack.swift
```diff
@@ -26,7 +26,7 @@ final class IOAudioMixerByMultiTrack: IOAudioMixerConvertible {
             }
             for id in tracks.keys {
                 buffers[id] = .init(outputFormat)
-                tracks[id] = .init(id: id, outputFormat: outputFormat)
+                tracks[id] = .init(id: id, outputFormat: outputFormat, delegate: self)
```
shogo4405, line 29: [IMO]
- I remember encountering a bug in some version of Swift where specifying a delegate in the constructor in a Release build would cause the weak reference to become nil. I thought it would be better to set the delegate in this line.
- `tracks[id]?.delegate = self`

levs42: Thanks, updated

PR #1501, opened 2024-06-28
Fix buffer rotation creation in Screencast example 

levs42:
## Description & motivation

This PR fixes initialization of the buffer rotator in Screencast example. 

## Type of change


shogo4405 (changes requested):
Simply enabling it might make it look buggy to first-time users. The default is set to landscape and trim mode, so when starting a vertical broadcast, the top and bottom get cut off. I think the following measures are necessary:

1. Set videoSettings.scaleMode to letterBox.
2. Allow it to be turned on/off with a flag or similar setting.

levs42:
Updated, please take a look

shogo4405 (approved):
👍

PR #1552, opened 2024-08-28
Add custom screen object to render effects on top of the stream

levs42:
## Description & motivation
* Feature: add dedicated video effects screen object

This PR introduces a new screen object that allows to render effects on top of the other screen objects. By default `IOVideoMixer` uses `videoTrackScreenObject` to render effects. As result, multicam covers the effects during the rendering. Having custom object allows to add it as a screen child and display extra elements on top of the stream. 

## Type of change

## Screenshots:
<details>
  <summary>Spoiler</summary>

Screenshot 2024-08-28 at 3 15 06 PM

  
</details>

### DIFF
--- a/Sources/Screen/ScreenObject.swift
+++ b/Sources/Screen/ScreenObject.swift
@@ -300,8 +300,13 @@ public final class VideoTrackScreenObject: ScreenObject, ChromaKeyProcessable {
 }
 
 /// An object that manages effects rendering.
-public final class EffectScreenObject: ScreenObject {
+public final class CIImageScreenObject: ScreenObject {
     private var effects: [VideoEffect] = .init()
+    private var image: CIImage?
+
+    public var hasEffects: Bool {
+        !effects.isEmpty
+    }
 
     /// Registers a video effect.
     public func registerVideoEffect(_ effect: VideoEffect) -> Bool {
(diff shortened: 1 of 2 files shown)


shogo4405 (changes requested):
From a maintenance perspective, I want to avoid increasing the number of ScreenObjects optimized for specific needs.

If place it on the HaishinKit side, I would like to make it more general so that other users can also utilize it easily. It can be accepted with the following changes:

Allow ImageScreenObject to be set from the CIImage library side since there is currently no CIImage-based ImageScreenObject.
Rename it to CIImageScreenObject."

levs42:
I've updated the PR, please take a look. I also added an extra check to `ScreenRendererByCPU` to avoid the situation where screen object was rendered for some time, than it became invisible, but the rendered still has an image in the `images` array and renders this cached image.

shogo4405:
I believe this is a class that should be placed in the app layer as a unique extension on the app side. There is no benefit in placing it on the HK side, so please position it in the app layer.

levs42:
Sounds good. How about `ScreenRendererByCPU` cached image issue, is the extra check something that can be merged?

shogo4405:
I cannot make a decision because there is no material (such as tests or reproducible code) to judge from. Let’s close this as a PR. When you have time, please provide reproducible code and create an issue. I will take care of the fix on my end.

From a design perspective, that check should be on the ScreenObjectContainer side, not on the ScreenRenderer side. https://github.com/shogo4405/HaishinKit.swift/blob/4adecd1a89ecde9bf873fe28911c17a0c93aa4ec/Sources/Screen/ScreenObjectContainer.swift#L60-L7

If I place it on the ScreenRenderer side, it could potentially cause other issues, or it might not. That’s why I’m being cautious.

PR #1561, opened 2024-09-08
Add the movieFragmentInterval value to the recording feature.

shogo4405:
## Description & motivation
- It possible to specify the movieFragmentInterval to recover from crashes or unintended task termination.

## Type of change

### DIFF
--- a/Sources/IO/IOStreamRecorder.swift
+++ b/Sources/IO/IOStreamRecorder.swift
@@ -44,6 +44,14 @@ public final class IOStreamRecorder {
     public var fileName: String?
     /// The running indicies whether recording or not.
     public private(set) var isRunning: Atomic<Bool> = .init(false)
+    /// Specifies the movie fragment interval in sec. This value allows the file to be written continuously, so the file will remain even if the app crashes or is forcefully terminated. A value of 10 seconds or more is recommended.
+    public var movieFragmentInterval: Double? {
+        didSet {
+            if let movieFragmentInterval {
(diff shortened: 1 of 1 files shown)


mkrn:
Thank you for adding this! 
I'm looking at the documentation 
https://developer.apple.com/documentation/avfoundation/avcapturemoviefileoutput/1387146-moviefragmentinterval
and it looks like it is 10s by default.. but the files are still sometimes unusable with crash.. was this feature disabled before?

shogo4405:
I have never implemented this before, so it will be a new implementation.

There are APIs using iOS that include [AVAssetWriter](https://developer.apple.com/documentation/avfoundation/avassetwriter) and [AVCaptureMovieFileOutput](https://developer.apple.com/documentation/avfoundation/avcapturemoviefileoutput). HaishinKit uses [AVAssetWriter](https://developer.apple.com/documentation/avfoundation/avassetwriter). 

In the case of AVAssetWriter, the default value for this is set to invalid.
https://developer.apple.com/documentation/avfoundation/avassetwriter/1387469-moviefragmentinterval

shogo4405:
I see, so it's unstable when it crashes. I was testing with Task Kill.
It seems like further research is needed, but for now, I believe this is the right specification and it will improve from here. I'll merge it for the time being.

mkrn:
Thank you this worked! It could be even 10.0s by default 
Thank you for pointing out that it was invalid in AVAssetWriter.

PR #1612, opened 2024-10-29
Added validation at the start of recording.

shogo4405:
## Description & motivation
- Added validation when a file is specified at the start of recording. An exception will be thrown if the file already exists.
  - https://github.com/shogo4405/HaishinKit.swift/pull/1611

## Type of change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### DIFF
--- a/Sources/HKStream/HKStreamRecorder.swift
+++ b/Sources/HKStream/HKStreamRecorder.swift
@@ -91,13 +91,15 @@ public actor HKStreamRecorder {
     public private(set) var movieFragmentInterval: Double?
     public private(set) var videoTrackId: UInt8? = UInt8.max
     public private(set) var audioTrackId: UInt8? = UInt8.max
-    #if os(iOS)
-    public private(set) lazy var moviesDirectory: URL = {
-        URL(fileURLWithPath: NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0])
-    }()
-    #else
+    #if os(macOS) && !targetEnvironment(macCatalyst)
+    /// The default file save location.
     public private(set) var moviesDirectory: URL = {
         URL(fileURLWithPath: NSSearchPathForDirectoriesInDomains(.moviesDirectory, .userDomainMask, true)[0])
+    }()
+    #else
+    /// The default file save location.
+    public private(set) lazy var moviesDirectory: URL = {
+        URL(fileURLWithPath: NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true)[0])
     }()
     #endif
     private var isReadyForStartWriting: Bool {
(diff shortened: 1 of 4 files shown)


Sources/HKStream/HKStreamRecorder.swift
```diff
@@ -109,21 +130,44 @@ public actor HKStreamRecorder {
...
+                outputURL = moviesDirectory.appendingPathComponent(file.absoluteString)
+            }
+            if file.pathExtension == "" {
+                outputURL = outputURL.appendingPathExtension(Self.defaultPathExtension)
+            }
+        } else {
+            outputURL = moviesDirectory.appendingPathComponent(UUID().uuidString).appendingPathExtension(Self.defaultPathExtension)
+        }
+
+        if FileManager.default.fileExists(atPath: outputURL.absoluteString) {
```
vkornel97, line 155: This should be `outputURL.path`.

Sources/HKStream/HKStreamRecorder.swift
```diff
@@ -109,21 +130,44 @@ public actor HKStreamRecorder {
...
+    ///   - settings: Settings for recording.
+    public func startRecording(_ file: URL? = nil, settings: [AVMediaType: [String: any Sendable]] = HKStreamRecorder.defaultSettings) async throws {
         guard !isRecording else {
             throw Error.invalidState
         }
 
-        self.fileName = fileName ?? UUID().uuidString
-        self.settings = settings
+        var outputURL: URL
+        if let file {
```
vkornel97, line 142: If you pass an URL that represents a file without a path extension (e.g. `file:///var/mobile/Containers/Data/Application/7910F2B5-9C52-42CE-87D0-20202839EB58/Documents/Demo` ) this logic would result in something like this: `file:///var/mobile/Containers/Data/Application/7910F2B5-9C52-42CE-87D0-20202839EB58/Documents/file:///var/mobile/Containers/Data/Application/7910F2B5-9C52-42CE-87D0-20202839EB58/Documents/Demo.mp4`, which is an incorrect file path.

A more future-proof suggestion to the file parameter and the outputURL handling:
  ```
      if let file {
            if file.hasDirectoryPath {
                outputURL = file.appendingPathComponent(UUID().uuidString)
            } else {
                outputURL = file
            }
            if outputURL.pathExtension == "" {
                outputURL = outputURL.appendingPathExtension(Self.defaultPathExtension)
            }
        } else {
            outputURL = moviesDirectory.appendingPathComponent(UUID().uuidString).appendingPathExtension(Self.defaultPathExtension)
        }

```

Tested with:

- URL representing a directory.
- URL representing a file without extension
- URL representing a file with extension
- nil URL

vkornel97 (changes requested):

Sources/HKStream/HKStreamRecorder.swift
shogo4405: Thank you for testing it. I’ve also added some test cases on my end. I’m thinking of adding the documentation and then merging.

PR #1621, opened 2024-11-09
Fix sending in SRT listener mode

caseyching:
## Description & motivation

This fixes a bug when using SRT in listener mode -- when trying to publish a stream, no data is actually sent. With this change, publication works for me.

## Type of change
- [x ] Bug fix (non-breaking change which fixes an issue)

## Screenshots:
N/A


shogo4405:
Thank you for your PR. There was a conflict, so I went ahead and fixed it on my end.

PR #1692, opened 2025-03-15
feat(Project): fix retain

sreshetnyak:
- fix retain for publish flow
- store async stream tasks and cancel it when disconnect

- it prevents frame duplication errors when, after reconnecting, we get a stream delay

- i have not checked the play stream flow, but I think it has the same problem with the async stream

test flow:

    private var connection: SRTConnection?
    private var stream: SRTStream?
    private var mixer: MediaMixer?

    func start() {
        let connection = SRTConnection()
        self.stream = SRTStream(connection: connection)
        self.mixer = MediaMixer(useManualCapture: true)
        self.connection = connection
        Task {
            do {
                guard let stream else { return }
                await mixer?.startRunning()
                await mixer?.addOutput(stream)
                try await connection.connect(URL(string: url))
                await stream.publish(name)
                _isRunning = true
            } catch {
                _isRunning = false
            }
        }
    }

    func stop() {
        Task {
            await mixer?.stopRunning()
            await connection?.close()
            await stream?.close()

            self.mixer = nil
            self.stream = nil
            self.connection = nil
            _isRunning = false
        }
    }

no we have proper object deinit

⚔️ deinit -> SRTHaishinKit.SRTSocket
⚔️ deinit -> SRTHaishinKit.SRTStream
⚔️ deinit -> SRTHaishinKit.SRTConnection

### DIFF
--- a/SRTHaishinKit/Sources/SRT/SRTConnection.swift
+++ b/SRTHaishinKit/Sources/SRT/SRTConnection.swift
@@ -23,9 +23,14 @@ public actor SRTConnection: NetworkConnection {
     /// This instance connect to server(true) or not(false)
     @Published public private(set) var connected = false
 
+    private var tasks = Set<Task<Void, Never>>()
+    
     private var socket: SRTSocket? {
         didSet {
-            Task {
+            tasks.forEach { $0.cancel() }
+            tasks.removeAll()
+            
+            let monitor = Task {
                 guard let socket else {
                     return
                 }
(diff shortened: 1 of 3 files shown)


shogo4405:
I believe there are two possible approaches:
1. Cancelling on the Task side.
1. Stopping on the AsyncStream.Continuation? side.

HaishinKit follows the second approach, and I plan to make adjustments using that method. I'll address this when I find some time, so please wait for a bit. The PR was helpful as a reference. Thank you.

shogo4405:
This improvement has been addressed in the referenced PR. I'll close this PR. 
https://github.com/HaishinKit/HaishinKit.swift/tree/feature/fix-srt-connetion-memory-leak

PR #1699, opened 2025-04-11
Changing code access protection to RTMPSocket.Error, to make it available for better client handling

peterpaulis:
## Description & motivation

Currently the RTMPSocket.Error cannot be properly handled by the client, due to its internal protection, this change makes the client able to react and handle the error more gracefully...

for example to extract/handle the core error, or provide custom UI messaging and handling

```
        if let e = rtmpError as? RTMPSocket.Error {
            switch e {
            case .connectionNotEstablished(let error):
                if let error {
                    //...
                }
            default:
                break
            }
        }
```

Follow up to https://github.com/HaishinKit/HaishinKit.swift/discussions/1698

## Type of change
- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/HaishinKit/Sources/RTMP/RTMPSocket.swift
+++ b/HaishinKit/Sources/RTMP/RTMPSocket.swift
@@ -1,10 +1,10 @@
 import Foundation
 import Network
 
-final actor RTMPSocket {
+public final actor RTMPSocket {
     static let defaultWindowSizeC = Int(UInt8.max)
 
-    enum Error: Swift.Error {
+    public enum Error: Swift.Error {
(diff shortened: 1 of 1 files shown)


shogo4405 (changes requested):
I understand that RTMPConnection.Error is a higher-level error of RTMPSocket.Error and can be handled accordingly. There are no plans to make RTMPSocket.Error public.

peterpaulis:
so there RTMPSocket.Error should than be moved into a public definition eg. moved out of the inaccessible class, as it is received by the client

PR #1700, opened 2025-04-11
Changing AVAudioPCMBuffer+Extension.swift access from internal to public

peterpaulis:
## Description & motivation

This is a very handy utility for general use, when making and handling custom output

```
    nonisolated public func mixer(_ mixer: MediaMixer, didOutput buffer: AVAudioPCMBuffer, when: AVAudioTime) {
        guard let sampleBuffer = buffer.makeSampleBuffer(when) else {
            return
        }
        Task {
            coreRecorder.externalCaptureDidOutputSampleBuffer(sampleBuffer, isVideo: false)
        }
    }
```

## Type of change
- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/HaishinKit/Sources/Extension/AVAudioPCMBuffer+Extension.swift
+++ b/HaishinKit/Sources/Extension/AVAudioPCMBuffer+Extension.swift
@@ -1,7 +1,7 @@
 import Accelerate
 import AVFoundation
 
-extension AVAudioPCMBuffer {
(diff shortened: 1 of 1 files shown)


shogo4405 (changes requested):
I understand that it's convenient, but I don't intend to make public the extensions to Apple-provided classes that are extended within the framework.

PR #1705, opened 2025-05-02
Fixed for multi channel mixer works well with Xcode16.3

mocona:
## Description & motivation

First of all, Thank you so much about you and your frameworks.

I'm using multi channel mixer mode from 1.8.x

It works well until Xcode 16.2, but it occurs runtime error with Xcode 16.3
(Seems MixerNode and OutputNode released so fast than expected - I assumed some kind of Swift runtime bug so I posted to [swift.org forums](https://forums.swift.org/t/different-instance-life-cycle-with-inout-between-xcode16-2-and-16-3/79631) )

To occur, just change multiTrackAudioMixingEnabled flags to true at 27th line in IngestViewController with Xcode 16.3

I fixed MixerNode description values as local value for inout arguments, to avoid runtime BAD_ACCESS_ERROR.

## Type of change
- Bug fix (non-breaking change which fixes an issue)

I hope it will be helpful.

### DIFF
--- a/Examples/iOS/IngestViewController.swift
+++ b/Examples/iOS/IngestViewController.swift
@@ -24,7 +24,7 @@ final class IngestViewController: UIViewController {
     private var retryCount: Int = 0
     private var preferedStereo = false
     private let netStreamSwitcher: HKStreamSwitcher = .init()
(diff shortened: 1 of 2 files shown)


shogo4405:
I was able to confirm the crash on my end as well. Thank you.

PR #1711, opened 2025-05-15
Make cgImage in ImageScreenObject thread-safe using a DispatchQueue

peterpaulis:
The cgImage property in ImageScreenObject is a public variable and can be updated at runtime. In scenarios where it’s frequently modified — such as updating a timestamp image every second (even from a Task { @ScreenActor in }) — thread safety issues can arise, leading to crashes. This change resolves the issue by ensuring cgImage access is thread-safe via a dedicated DispatchQueue.

## Type of change
- [X] Bug fix (non-breaking change which fixes an issue)


peterpaulis:
There seems to be another issue related to 60 FPS related to crashing, will try to write a bug report

PR #1736, opened 2025-06-30
Fix: Check of input buffer capacity, not the current length.

mstyura:
The input buffer can store appended buffer if it's capacity greater than or equal of amount of input. This reduces amount of allocations of audio buffers when appended buffers frequently changes `frameLength`.

## Description & motivation

The appended buffer could be written into the `inputBuffer` in case of `inputBuffer` has enough capacity, it does not matter how many samples were presented in the buffer before (aka `frameLength`).

## Type of change
- [ ] Enhancement (non-breaking change which fixes an issue)

### DIFF
--- a/HaishinKit/Sources/Mixer/AudioRingBuffer.swift
+++ b/HaishinKit/Sources/Mixer/AudioRingBuffer.swift
@@ -21,6 +21,11 @@ final class AudioRingBuffer {
     private var sampleTime: AVAudioFramePosition = 0
     private var inputFormat: AVAudioFormat
     private var inputBuffer: AVAudioPCMBuffer
+#if DEBUG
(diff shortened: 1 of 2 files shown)


shogo4405 (changes requested):
In the test cases I envision, this error condition does not occur, so there is no benefit in accepting the PR.
Since I'm concerned about potential regressions, adding test cases would increase the likelihood of acceptance.

mstyura:
I supposed I need to test that the implementation side effect (excessive allocation) is gone, that's why I needed to expose internal buffer - to check implementation behaviour. Regarding the check of `input -> output` I believe the existing tests already covers it, I'm not sure what new I can add here.

shogo4405:
If that's the case, then there's no particular reason to accept the PR, so let's close this discussion.

PR #1864, opened 2025-12-30
Fix: AudioRingBuffer wrap-around crash

devjoonn:
## Description & motivation
`AudioRingBuffer.append(_:offset:)` could crash with `EXC_BAD_ACCESS` when the ring wrapped, because recursive calls restarted at `numSamples` instead of accumulating the original offset.

- Added a single guard ensuring `offset`/`head` stay within `[0, frameLength)`.
- When the buffer wraps, we now recurse with `nextOffset = offset + numSamples` so the copy resumes from the correct position.

## Type of change

### DIFF
--- a/HaishinKit/Sources/Mixer/AudioRingBuffer.swift
+++ b/HaishinKit/Sources/Mixer/AudioRingBuffer.swift
@@ -193,21 +193,13 @@ final class AudioRingBuffer {
         let numSamples = min(Int(audioPCMBuffer.frameLength) - offset, Int(outputBuffer.frameLength) - head)
         if inputFormat.isInterleaved {
             let channelCount = Int(inputFormat.channelCount)
-            let outputOffset = head * channelCount
-            let inputOffset = offset * channelCount
             switch inputFormat.commonFormat {
             case .pcmFormatInt16:
-                guard let outputPtr = outputBuffer.int16ChannelData?[0],
-                      let inputPtr = audioPCMBuffer.int16ChannelData?[0] else { return }
-                memcpy(outputPtr.advanced(by: outputOffset), inputPtr.advanced(by: inputOffset), numSamples * channelCount * 2)
+                memcpy(outputBuffer.int16ChannelData?[0].advanced(by: head * channelCount), audioPCMBuffer.int16ChannelData?[0].advanced(by: offset * channelCount), numSamples * channelCount * 2)
             case .pcmFormatInt32:
-                guard let outputPtr = outputBuffer.int32ChannelData?[0],
-                      let inputPtr = audioPCMBuffer.int32ChannelData?[0] else { return }
-                memcpy(outputPtr.advanced(by: outputOffset), inputPtr.advanced(by: inputOffset), numSamples * channelCount * 4)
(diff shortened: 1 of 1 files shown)


shogo4405 (changes requested):
In what kind of use cases would this type of issue occur?

Also, since there are no test cases, it is not possible to determine whether this change is good or bad, or whether it introduces other issues, so I cannot accept it.

devjoonn:
Thank you for the quick review.
Please let me know if there is anything else that needs to be reviewed, and I will provide additional information.

## Issue Reproduction Cases
This issue occurs when switching cameras or repeatedly toggling mute during streaming.

### 1. Camera Switching

```swift
try await rtmpMixer.attachAudio(AVCaptureDevice.default(for: AVMediaType.audio))
try await rtmpMixer.attachVideo(camera) { videoUnit in
    videoUnit.isVideoMirrored = !isBackCam
    videoUnit.preferredVideoStabilizationMode = .off
    try videoUnit.setFrameRate(30)
}

await rtmpMixer.startRunning()
await rtmpMixer.startCapturing()
```

### 2. Mute Toggle

```swift
await rtmpMixer.setAudioMixerSettings(.init(isMuted: needToMute))
```

## Crash Image
스크린샷 2025-12-30 오전 11 24 17


## Crash Call Stack
```
thread #54, queue = 'com.haishinkit.HaishinKit.AudioCaptureUnit.lock', stop reason = EXC_BAD_ACCESS (code=2, address=0x17063bfe0)
  * frame #0: AudioRingBuffer.append(audioPCMBuffer=0x0000000300a48600, offset=16384) at AudioRingBuffer.swift:189:76
    frame #1: AudioRingBuffer.append(audioPCMBuffer=0x0000000300a48600, offset=16384) at AudioRingBuffer.swift:221:17 (recursive call)
    frame #237: AudioRingBuffer.append(audioPCMBuffer=0x0000000300a48600, offset=0) at AudioRingBuffer.swift:221:17
    frame #238: AudioRingBuffer.append(sampleBuffer=0x132545f80) at AudioRingBuffer.swift:83:9
    frame #239: AudioMixerTrack.append(sampleBuffer=0x132545f80) at AudioMixerTrack.swift:67:21
    frame #240: AudioMixerBySingleTrack.append(track=0, buffer=0x132545f80) at AudioMixerBySingleTrack.swift:45:21
    frame #242: AudioDeviceUnitDataOutput.captureOutput(output=..., sampleBuffer=0x132545f80, connection=...) at AudioDeviceUnit.swift:58:20
    frame #244: AVCaptureAudioDataOutput._handleSampleBufferEventForSampleBuffer: + 156
```

shogo4405:
I believe this crash occurs because the data exceeds the maximum buffer size (16,384) that the AudioRingBuffer can store.

As a possible fix, I think it would be better to skip processing instead of calling
append(_ audioPCMBuffer: AVAudioPCMBuffer, offset: Int = 0).

To better understand the situation, I would like to see the output of the CMSampleBuffer at the time of the crash so we can know how much data was actually being buffered.

https://github.com/HaishinKit/HaishinKit.swift/blob/47ccc4bf5bbbdbc9a444bb32c8ace4a8e4a9c23a/HaishinKit/Sources/Mixer/AudioRingBuffer.swift#L44-L52

devjoonn:
@shogo4405 Confirmed that no crash occurs. #1867 
The guard logic for sample buffers exceeding the ring buffer size works as expected. 
Thank you.

PR #1937, opened 2026-08-10
Update MediaMixer.swift

peterpaulis:
Fixed issue on loosing audio on longer background state

## Description & motivation

Encountered issue, where the audio was being lost after being in background for longer

## Type of change
- [ X] Bug fix (non-breaking change which fixes an issue)


shogo4405:
I think enabling microphone access while the app is running in the background has actually made it unavailable. I’m going to close this issue.

peterpaulis:
@shogo4405  Are you sure that the missing suspending and resuming of audio while changing background state its not an issue?

PR #1940, opened 2026-09-08
Guard VideoMixer.inputFormats against concurrent mutation

EnderTang:
A data race on `VideoMixer.inputFormats` crashes the app on camera attach/detach.

`inputFormats` is a plain `Dictionary` mutated from two queues with no synchronisation:

| | writer | queue |
|---|---|---|
| `append(_:sampleBuffer:)` | `VideoCaptureUnitDataOutput.captureOutput` | `AVCaptureVideoDataOutput` sample-buffer queue, **every frame** |
| `reset(_:)` | `VideoCaptureUnit.attachVideo` | `VideoCaptureUnit.lockQueue` |

The two overlap **by construction** rather than by bad luck: `attachVideo` performs the detach inside

```swift
try session.configuration { _ in
    session.detachCapture(devices[track])
    videoMixer.reset(track)
```

and `configuration { }` batches the change without stopping the running session — so frames are still being delivered into `append()` while `reset()` removes the key.

Swift `Dictionary` is not thread-safe, and concurrent mutation corrupts its storage. In production this shows up as a fatal, unhandled exception reading a freed slot, on the camera toggle:

```
-[__NSTaggedDate objectForKey:]: unrecognized selector sent to instance 0x8000000000000000
  Dictionary._Variant.removeValue  <-  VideoMixer.reset  <-  attachVideo
```

**The fix** is an `NSLock` around the three access sites, rather than a queue hop, because `append()` runs per frame and should not pay a dispatch. The critical section is a single dictionary access and never calls out — the delegate callback is deliberately kept *outside* it, since a re-entrant delegate would otherwise deadlock.

The 1.x line is affected identically (`IOVideoMixer.inputFormats`, with `IOVideoUnit.attachCamera` using the same `session.configuration { }` block); happy to open the equivalent patch there if 1.x is still taking fixes.


EnderTang:
Closing this — opened prematurely on my side, before it had been reviewed internally. Apologies for the noise. The underlying report stands and I'll re-open properly if it's useful.

